PR #353, opened 2015-09-16
Use HTTP_X_FORWARDED_HOST if available

holzgeist:
This enables piwigo to be run e.g. in a docker container behind a proxy.
If HTTP_X_FORWARDED_HOST is not set, HTTP_HOST is used as usual

### DIFF
--- a/include/functions_url.inc.php
+++ b/include/functions_url.inc.php
@@ -69,14 +69,14 @@ function get_absolute_root_url($with_scheme=true)
     else
     {
       $url .= $_SERVER['HTTP_HOST'];
-    }
-    if ( (!$is_https && $_SERVER['SERVER_PORT'] != 80)
-          ||($is_https && $_SERVER['SERVER_PORT'] != 443))
-    {
-      $url_port = ':'.$_SERVER['SERVER_PORT'];
-      if (strrchr($url, ':') != $url_port)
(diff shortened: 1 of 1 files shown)


plegall:
I think it may solve an issue I encountered today: I have a Piwigo running on Apache (port 8000) behind an Nginx reverse proxy (port 80). The url returned by Piwigo is host:8000 while I don't want it to be :8000 ... I will try that.

plegall:
Actually it doesn't solve my problem :-/ It's because I have a $_SERVER[SERVER_PORT] = 8000. Don't you have the same problem?

holzgeist:
Yes, actually I had the same problem without knowing it.. :/
I have an apache server on port 80 running within a docker container. The traffic is forwarded to some high port on the host which is used by the proxy to communicate with piwigo (http://localhost:high_port).
Piwigo returned host:high_port (which is wrong) but as docker opens the forwarded port in the firewall on the host, it still worked.

But assuming that the proxy fills the 'Http-X-Forwarded-Host' header correctly, we could take it as it is if found, and use SERVER_PORT otherwise.
In your case it is still different, as nginx changes the 'Host' header directly without setting the 'Http-X-Forwarded-Host' and so we don't know that there is a proxy at all :(
You could reconfigure nginx to set 'Http-X-Forwarded-Host': proxy_set_header X-Forwarded-Host $host:$server_port;

PR #371, opened 2015-11-08
fix #367 - do not specialize numeric tag to 'id-tag' URL

### LINKED ISSUE #367: tag_url_style=tag fails on numeric tags
I have

`$conf['tag_url_style'] = 'tag';`

in my local config, which generally works, but fails on numeric tags where it falls back to id-tag, for example instead of "2015"  the tag "88-2015" is in the underlying URL. I wouldn't mind that much if clicking such would actually lead to the group of images tagged with "2015", but instead it gives "Page not found" and then redirects to the tags.php page. I reckon that probably the id is prefixed because of an alleged ambiguity of numeric tag name and numeric tag id, but then it should actually work.. Note that removing the "id-" part from the URL so that only the actual numeric tag is left works fine and the corresponding tagged images are displayed.

erAck:
Apparently there is no reason in make_section_in_url() to "escape" a
numeric URL for a tags section if tag_url_style is 'tag'. In
parse_section_url() tag_url_style is checked, and if 'tag' the URL token
is used.

### DIFF
--- a/include/functions_url.inc.php
+++ b/include/functions_url.inc.php
@@ -379,7 +379,7 @@ function make_section_in_url($params)
             $section_string.= '/'.$tag['id'];
             break;
           case 'tag':
-            if (isset($tag['url_name']) and !is_numeric($tag['url_name']) )
+            if (isset($tag['url_name']))
             {
               $section_string.= '/'.$tag['url_name'];
               break;


mistic100:
@plg @modus75 is it ok for you ?

plegall:
For me it's fine, I have tested: without this change Piwigo generates broken links for numeric tags, with this change we have working links.

Issue #324, opened 2015-07-25
[image processing] missing thumbnail for TIFF on upload form


plegall:
**Reported by plg on 21 Apr 2015 11:15**

**Version:** 2.7.4

When upload a TIFF we only get a &quot;missing thumbnail&quot; icon instead of the appropriate thumbnail.

[Piwigo Bugtracker #3219](http://piwigo.org/bugs/view.php?id=3219)

Philippe-M:
partial correction. In file admin/include/image.class.php in write function in image_ext_magick

  function write($destination_filepath)
  {
  	global $conf;
  	
    $this->add_command('interlace', 'line'); // progressive rendering
    // use 4:2:2 chroma subsampling (reduce file size by 20-30% with "almost" no human perception)
    //
    // option deactivated for Piwigo 2.4.1, it doesn't work fo old versions
    // of ImageMagick, see bug:2672. To reactivate once we have a better way
    // to detect IM version and when we know which version supports this
    // option
    //
    if (version_compare(pwg_image::$ext_imagick_version, '6.6') > 0)
    {
      $this->add_command('sampling-factor', '4:2:2' );
    }

    $exec = $this->imagickdir.'convert';
    $exec .= ' "'.realpath($this->source_filepath).'"';

    foreach ($this->commands as $command => $params)
    {
      $exec .= ' -'.$command;
      if (!empty($params))
      {
        $exec .= ' '.$params;
      }
    }    
    
    $dest = pathinfo($destination_filepath);        
    if ($dest['extension'] == 'tif') {
    	$exec .= ' "'.realpath($dest['dirname']).'/'.$dest['filename'].'.'.$conf['tiff_representative_ext'].'" 2>&1';
    } else {
    	$exec .= ' "'.realpath($dest['dirname']).'/'.$dest['basename'].'" 2>&1';
    }
    
    @exec($exec, $returnarray);    

    if (function_exists('ilog')) ilog($exec);
    if (is_array($returnarray) && (count($returnarray)>0) )
    {
      if (function_exists('ilog')) ilog('ERROR', $returnarray);
      foreach($returnarray as $line)
        trigger_error($line, E_USER_WARNING);
    }
    return is_array($returnarray);
  }
}

plegall:
I just made a new test (on current git master) and the problem does not occur. I have to make so more tests.

plegall:
[...]

PR #374, opened 2015-11-16
fixe #352 Add upload and generate thumbnails for EPS files

Philippe-M:
Add in local/config/config.inc.php

$conf['postscript_extension'] = array('eps', 'EPS');
$conf['postscript_representative_ext'] = 'jpg';


plegall:
@Philippe-M I see that it's the same as for TIFF, but I still would like some details. I don't understand why you changed these 2 files. As I said on #324 I don't find any problem with TIFF files!

PR #387, opened 2015-12-29
Add better support for HTTPS Piwigo installations

lcorbasson:
- in the WS self-documentation

### DIFF
--- a/tools/ws.htm
+++ b/tools/ws.htm
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <title>Piwigo web API (web-services) explorer</title>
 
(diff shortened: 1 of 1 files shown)


modus75:
Better use href and src="//... rather than hard-coded http or https. No?

plegall:
I agree with @modus75 I think the src="// was created exactly for that. Can you make the test @lcorbasson and update your branch for the pull-request?

mistic100:
Resolved with #391

lcorbasson:
Thanks @modus75, @plegall, @mistic100!

PR #391, opened 2015-12-30
Modify /tools by removing http: from URLs which support https

thunderrabbit:
Thanks to @lcorbasson for attempting this originally: https://github.com/Piwigo/Piwigo/pull/387


mistic100:
Thank you

PR #393, opened 2015-12-31
Feature/392 auth keys

plegall:


modus75:
Hello, If I'm not wrong the auto increment id in the db table is useless, Shouldn't the auth_key be the primary key ?
Happy new year :-)

plegall:
@modus75 just like image_format.format_id, it is not useless :-) I plan to add either a dedicated history tables for auth_keys usage OR add a history.auth_key_id column. In the past I have sometimes avoided to create such primary key, and then realized I needed them and had to add them afterwards!

Happy new year ;-)

include/functions_user.inc.php
```diff
@@ -1462,4 +1462,136 @@ function get_recent_photos_sql($db_field)
...
+  
+  $query = '
+SELECT
+    COUNT(*),
+    NOW(),
+    ADDDATE(NOW(), INTERVAL '.$conf['auth_key_duration'].' SECOND)
+  FROM '.USER_AUTH_KEYS_TABLE.'
+  WHERE auth_key = \''.$candidate.'\'
+;';
+  list($counter, $now, $expiration) = pwg_db_fetch_row(pwg_query($query));
```
mistic100: I would have detect collisions on insert error (if there is a unique key in the scheme) because collisions will be very very rare with 30 chars key.

include/functions_user.inc.php
```diff
@@ -1462,4 +1462,136 @@ function get_recent_photos_sql($db_field)
...
+    
+    single_insert(USER_AUTH_KEYS_TABLE, $key);
+
+    $key['auth_key_id'] = pwg_db_insert_id();
+    
+    return $key;
+  }
+  else
+  {
+    return create_user_auth_key($user_id);
```
mistic100: Isn't `$user_status` missing ?

PR #396, opened 2016-01-09
Japanese word for "author" depends on context

mathieumd:
作成者 (sakuseisha) is more generic than 撮影者 (satsueisha).
It's isn't perfect, though. The best way would be to have a different string depending on context:
- Photos' author -> 撮影者 (satsueisha)
- Comment's author -> 記入者 (kinyuusha)


mistic100:
Thank you, you can also contribute on our translation platform http://piwigo.org/translate (requires an account on the English forum).

Contact @ddtddt for permissions.

mistic100:
Cherry picked on translation branch https://github.com/Piwigo/Piwigo/commit/1cc56757083260a6570b434f1cc2e58919d992c1

Issue #397, opened 2016-01-09
[user manager] edit user in a popin


plegall:
... instead of an embedded form into users table.

PR #399, opened 2016-01-09
feature #397, edit user in a popin

plegall:


mistic100:
I close because included in #400

PR #400, opened 2016-01-09
Feature/398 refresh user list

plegall:


mistic100:
Why the CSS is not in a separated file ?

plegall:
because we already had some {html_style} in user_list.tpl so I didn't made "another way". I moved the new rules to theme.css

mistic100:
When I delete the user, the popin is not closed

Issue #359, opened 2015-10-08
database connection error with PHP 5.6


plegall:
http://piwigo.org/forum/viewtopic.php?id=25729

functions_mysqli.inc.php, line 65

$dbname = null;

must be changed to :

$dbname = '';

... or else we get a fatal error "Warning: mysqli::mysqli(): (42000/1102): Incorrect database name '0Oõ»áÈ ' in /home/magicphorn/www/piwigo/include/dblayer/functions_mysqli.inc.php on line 68"

plegall:
It seems that some PHP 5.6.8 and 5.6.6 don't have the bug.

PR #458, opened 2016-04-07
fixing #359 and defining connection early

Zvax:
slight changes to the way connection is defined for each case, mysqli is created directly within each ifs

removed the dbname = null and replaced it with the default value for mysqli construct

that's pretty much written in the code anyway, cheers

### DIFF
--- a/include/dblayer/functions_mysqli.inc.php
+++ b/include/dblayer/functions_mysqli.inc.php
@@ -48,23 +48,21 @@ define('DB_RANDOM_FUNCTION', 'RAND');
 function pwg_db_connect($host, $user, $password, $database)
 {
   global $mysqli;
-
-  $port = null;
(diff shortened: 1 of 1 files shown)


plegall:
Hi @Zvax, actually I don't see the point to write 3 "new mysqli()" statements instead of 1. For the first one, it's really "less explicit" than the original code in my opinion. I mean that with original code, I understand the $socket gets the value of $host and $host becomes null. With your code I fear it becomes less readable (ie, harder to maintain).

Zvax:
Hey there, thanks for following up.

I agree with the less explicit socket vs host part, I changed it back so that it stand obvious.

However, I do think that separating the three cases helps understand how it actually works. Before, I wasn't exactly sure what was happening why and for which connection case.

Cheers

Issue #467, opened 2016-04-22
Image sharpening produce black images with PHP 7.0 and GD


juillerat:
This bug has been reproduced on Centos 6.7 with PHP 7.0.4 and Ubuntu with php 7.0.5, both with GD library.
For example, generation of XXS thumbnails produce black images with default configuration. The problem lies in admin/includes/image.class.php in the get_sharpen_matrix function. Precisely, the normalized matrix will output bad images with the GD imageconvolution function, without throwing errors.
Changing the normalizing code from

<pre lang="javascript"><code>
for ($i=0; $i<3; $i++)
{
    $line = & $matrix[$i];
    for ($j=0; $j<3; $j++)
    {
      $line[$j] /= $norm;
    }
}
</code></pre>

To

<pre lang="javascript"><code>
for ($i=0; $i<3; $i++)
{
    for ($j=0; $j<3; $j++)
    {
      $matrix[$i][$j] /= $norm;
    }
}
</code></pre>

resolve the problem.

olaulau:
I can confirm that this is an old annoying bug, since the app was told to be PHP7 compatible since 2.8.0, and it globally seems to work well with PHP7.

PR #468, opened 2016-04-22
fixing #467 and correcting image sharpening on PHP 7.0 with GD

juillerat:
Fix the normalization of the convolution kernel for sharpening


plegall:
@modus75 can you review that please? Would that have side effects?

PR #474, opened 2016-05-03
Add AI picture upload

Philippe-M:
Add ai extension in

$conf['file_ext']

active

$conf['enable_formats'] = true;
$conf['upload_form_all_types'] = true;


plegall:
Hi @Philippe-M 

Why don't you create a plugin for that? "AI file type" for example. I don't think it's smart to include all file types in Piwigo core.

PR #477, opened 2016-05-06
fix single_insert when a field is null

### LINKED ISSUE #449: Video upload fails
Uploading a video clip to Piwigo 2.8.0 results in the following error:

```
**REQUEST**:
POST http://piwigo28.int/ws.php?method=pwg.images.add&format=json HTTP/1.1
Accept: */*
Cookie: pwg_id=8arvv47c2tha51v2n20cvpj2t4
Content-Type: application/x-www-form-urlencoded
Content-Length: 261
User-Agent: Adobe Photoshop Lightroom
Host: piwigo28.int
Pragma: no-cache

categories=2&file_sum=a73c256036477811d94521561c2d36ce&comment=&author=&tag_ids=1&pwg_token=56c47f9a677e25f99bfcf19fde4f8201&name=&original_sum=a73c256036477811d94521561c2d36ce&date_creation=2014-09-01%2022%3A49%3A37&level=0&resize=0&original_filename=-9035.mp4

**RESPONSE**:

HTTP/1.1 200 OK
Date: Sun, 03 Apr 2016 19:50:07 GMT
Server: Apache/2.2.22 (Win32) mod_ssl/2.2.22 OpenSSL/0.9.8t PHP/5.3.29
X-Powered-By: PHP/5.3.29
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
Pragma: no-cache
Content-Length: 584
Content-Type: text/plain; charset=utf-8

<pre><br />
<b>Warning</b>:  [mysql error 1366] Incorrect integer value: '' for column 'width' at row 1

INSERT INTO piwigo_images
  (file,name,date_available,path,filesize,width,height,md5sum,added_by,rotation,level)
  VALUES('-9035.mp4','-9035','2016-04-03 15:50:07','./upload/2016/04/03/20160403155007-a73c2560.mp4','2962','','','a73c256036477811d94521561c2d36ce','1','0','0') in <b>D:\WWW_HOME\piwigo28.com\include\dblayer\functions_mysqli.inc.php</b> on line <b>832</b><br />
</pre>{"stat":"ok","result":{"image_id":0,"url":"http:\/\/piwigo28.int\/picture.php?\/0\/category\/2"}}
```

dkm:
fix #449

Signed-off-by: Marc Poulhiès dkm@kataplop.net

### DIFF
--- a/include/dblayer/functions_mysql.inc.php
+++ b/include/dblayer/functions_mysql.inc.php
@@ -503,7 +503,7 @@ INSERT INTO '.$table_name.'
         $is_first = false;
       }
       
-      if ($value === '')
+      if ($value === '' || is_null($value))
       {
         $query .= 'NULL';
       }
--- a/include/dblayer/functions_mysqli.inc.php
+++ b/include/dblayer/functions_mysqli.inc.php
@@ -581,7 +581,7 @@ INSERT INTO '.$table_name.'
         $is_first = false;
       }
       
-      if ($value === '')
+      if ($value === '' || is_null($value))
       {
         $query .= 'NULL';
       }


plegall:
@dkm I fear the side effects of this, I'm going to test it as much as possible...

plegall:
I don't feel comfortable with the difference in mass_inserts and in single_insert (which already exists in Piwigo 2.8.1)

1) in mass_inserts, to get a NULL in MySQL, we have these conditions:

`!isset($insert[$dbfield]) or $insert[$dbfield] === ''`

2) in single_insert, to get the same NULL in MYSQL, we have this condition:

`$value === ''`

We should, at least, have the same conditions.

Your is_null, for a null value, would produce exactly the same result as !isset.

Actually, an even more better condition would simply be (for both functions):

```
if (empty($value)) // matches null or "" or not defined (in case of array value)
{
  $query .= 'NULL';
}
```

modus75:
I think empty(0) === true so it's not a good test

plegall:
You're right @modus75 

```
$var = 0;
var_dump(empty($var));
```

gives bool(true).

So... if we consider the "is_null" proposed by @dkm for single_insert is already covered by "isset" in mass_inserts, we can consider Piwigo will have the same behavior with mass_inserts and single_insert, once the PR will be applied. Do you agree @modus75 ?

PR #482, opened 2016-05-22
Fix layout for IE mobile 11+/Windows Phone 8.1+

tkuther:
IE mobile on Windows Phone reports itself with a way too large viewport size. This change hints IE to use the proper device width. Works with Windows Phone 8.1 upwards (IE11+) Windows Phone 8 would require an additional step

For before/after screenshots and discussion see http://piwigo.org/forum/viewtopic.php?id=26612

Reference: http://getbootstrap.com/getting-started/#support-ie10-width

### DIFF
--- a/themes/smartpocket/theme.css
+++ b/themes/smartpocket/theme.css
@@ -1,6 +1,8 @@
 @import "jquery.mobile.css";
 @import "photoswipe.css";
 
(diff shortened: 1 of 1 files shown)


tkuther:
It turned out that the issue was related to a problem on that user's instalation in someway. The fix here actually didn't have any effect at all. Closing...

PR #483, opened 2016-05-22
Honor X-Forwarded-Proto HTTP Header when set by a frontal reverse-proxy

mossroy:
This fixes a bug in batch resize feature when you try to run it from behind a reverse-proxy that handles SSL encryption, while the server that runs Piwigo serves unencrypted content :
client -> https://reverse-proxy -> http://piwigo_server

In this case, $_SERVER['HTTPS'] is not set on piwigo_server, so get_absolute_root_url would give http://reverse-proxy (instead of https://reverse-proxy)

But if you set the X-Forwarded-Proto header (see https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields) in the reverse-proxy, this patch handles this case properly.
This can be done on apache with this directive :
RequestHeader set X-Forwarded-Proto "https"

This patch is inspired by this ticket from Wordpress : https://core.trac.wordpress.org/ticket/15733
Please be aware that the patch has not been merged on Wordpress : they preferred to invite users to add some PHP code in their config to handle this manually (see http://codex.wordpress.org/Administration_Over_SSL) :
if (strpos($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') !== false)
       $_SERVER['HTTPS']='on';

The same method can be applied on Piwigo (by adding the same lines of code in local/config/config.inc.php, through the LocalFiles Editor plugin). But I would find it much better to include that in the code of Piwigo, as HTTP_X_FORWARDED_HOST is already handled.

### DIFF
--- a/include/functions_url.inc.php
+++ b/include/functions_url.inc.php
@@ -52,8 +52,10 @@ function get_absolute_root_url($with_scheme=true)
   if ($with_scheme)
   {
     $is_https = false;
-    if (isset($_SERVER['HTTPS']) &&
+    if ((isset($_SERVER['HTTPS']) &&
       ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
+    || (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) &&
+      strpos($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') !== false))
     {
       $is_https = true;
       $url .= 'https://';


modus75:
I disagree with implementing this as default. I think anyone can send any http header, so changing the SERVER variables in your local config.inc.php to reflect the server/port as seen by the client in a better solution.(you know for sure your architecture has a proxy)

mossroy:
I understand.
But I'm wondering why it is already handled by default for X-Forwarded-Host header (just a few lines below the ones I modified)?
Wouldn't it be more consistent to handle either both or none of these headers?

modus75:
You are right. Someone has added it. In my opinion wrongfully.

SpComb:
Only adjusting the `https://` protocol part based on the HTTP X-Forwarded-Proto header is not enough, as the `$_SERVER['SERVER_PORT']` will remain 80, leading to URLS of the form `https://host:80/` when behind a reverse proxy.

plegall:
The recommended solution for now is described by @mossroy on https://github.com/Piwigo/Piwigo/pull/1637#issuecomment-1166576255

bertrandgorge:
Hello, thanks for the fix ! I tested it on my setup (docker with a Traefik reverse proxy up front), and I confirm it fixes it.

plegall:
See https://github.com/Piwigo/Piwigo/commit/2e037205ffae545b324c4cbcdac1c986db1ba6e5

PR #495, opened 2016-06-06
fixes #429 upgrade message

### LINKED ISSUE #429: discover new feature button at the end of upgrade
Add core language string for Take a Tour button at the end of upgrade. Currently (Piwigo 2.8), we have

`str_replace('2.7', get_branch_from_version(PHPWG_VERSION), l10n('2_7_0_descrp'))`

which would be cleaner as a core language string "Discover what's new in Piwigo %s"

jeremydurand:


upgrade.php
```diff
@@ -466,22 +466,12 @@ function print_time($message)
...
-            'force_fallback'=>'en_UK',
-            )
-          );
-
         // we need the secret key for get_pwg_token()
         load_conf_from_db();
         
         $template->assign(
           array(
             // TODO find a better way to do that, with a core string in English
```
plegall: maybe we don't need this TODO anymore?

upgrade.php
```diff
@@ -466,22 +466,12 @@ function print_time($message)
...
-            'button_label' => str_replace('2.7', get_branch_from_version(PHPWG_VERSION), l10n('2_7_0_descrp')),
```
plegall: you removed this line with no replacement?

PR #510, opened 2016-07-27
Set correct Contenttype for response (fixes #406)

### LINKED ISSUE #406: API responses - the Content-Type header should be consistent with response format
I noticed that Piwigo (tested on version 2.7.4) supplies an incorrect Content-Type header in its API responses. It says text/plain when it should be *application/json*. 
When response is XML, the Content-Type needs to be *application/xml*.
Thank you.

5nafu:
I just updated the json_encoder to fix this issue.

### DIFF
--- a/include/ws_protocols/json_encoder.php
+++ b/include/ws_protocols/json_encoder.php
@@ -47,7 +47,7 @@ class PwgJsonEncoder extends PwgResponseEncoder
 
   function getContentType()
   {
-    return 'text/plain';
+    return 'text/json';
   }
 }
 


5nafu:
jQuery aparently throws an error if the data to parse is already an (json) object. I fixed this in the ws.htm file, but could somebody please check if the following occurrences for conflicts? Unfortunately I do not know there they are used:
```
admin/themes/default/template/cat_modify.tpl:        var data = jQuery.parseJSON(data);
admin/themes/default/template/cat_modify.tpl:        var data = jQuery.parseJSON(data);
admin/themes/default/template/photos_add_direct.tpl:        var data = jQuery.parseJSON(info.response);
admin/themes/default/template/user_list.tpl:        var data = jQuery.parseJSON(data);
admin/themes/default/template/user_list.tpl:        var data = jQuery.parseJSON(data);
admin/themes/default/template/user_list.tpl:        var data = jQuery.parseJSON(data);
admin/themes/default/template/user_list.tpl:        var data = jQuery.parseJSON(data);
```

plegall:
@5nafu actually I'm going to test (I'm very late on this, sorry) because I fear the side effect of this change...

PR #531, opened 2016-09-30
API call to get the user's favorite images

abrenoch:
I thought the naming of the API call and placement of the function (within pwg.php in oppose to a different file) made sense, but if anyone thinks there is a better option that is cool too!


plegall:
Hi @abrenoch 

I'm very late on this! Let's go forward if we want it into Piwigo 2.9 :-)

The idea is very good. We would eventually need a way to set/unset a photo as favorite for a user through API, but this is a good start to get the list. I think it would be better to change a few things on your PR:

- the method should be pwg.users.getFavorites
- function ws_users_getFavorites moved to include/ws_functions/pwg.users.php
- the method could take an optional parameter user_id, if set and if the current user is_admin() then we can use it instead of $user['id']. That would be optimal, but not required to merge the PR ;-)

abrenoch:
Just applied the noted revisions.. I'll submit a new pull request!

PR #538, opened 2016-10-14
Initial mysqli TLS/SSL support?

martinstefany:
Hi,

I've patched my Piwigo installation to support TLS/SSL db connectivity via mysqli, so I'm sending a patch for it.
TLS/SSL parameters are not exposed via installation webpage with this patch, they can be only edited after installation in file local/config/database.inc.php. Installer is patched to insert NULL values by default into local/config/database.inc.php, but upgrade path somehow also needs a way to insert these into the file if they are missing. Or should these be set in include/config_default.php and local/config/database.inc.php will be just override?

I know this is not ready to merge, but please guide me on resolving the missing pieces. My skills with HTML and PHP are minimal.

Thank you in advance.
Martin

### DIFF
--- a/admin/include/functions_upgrade.php
+++ b/admin/include/functions_upgrade.php
@@ -311,7 +311,10 @@ function upgrade_db_connect()
   try
   {
     pwg_db_connect($conf['db_host'], $conf['db_user'],
-                   $conf['db_password'], $conf['db_base']);
+                   $conf['db_password'], $conf['db_base'],
+                   $conf['db_sslkey'], $conf['db_sslcert'],
+                   $conf['db_sslca'], $conf['db_sslcapath'],
+                   $conf['db_sslcipher']);
     pwg_db_check_version();
   }
   catch (Exception $e)
--- a/i.php
+++ b/i.php
@@ -393,7 +393,10 @@ include_once( PHPWG_ROOT_PATH .'/include/derivative_std_params.inc.php');
 try
 {
   pwg_db_connect($conf['db_host'], $conf['db_user'],
-                 $conf['db_password'], $conf['db_base']);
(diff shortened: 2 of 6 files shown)


plegall:
Hi @martinstefany 

Thank you for your PR. Indeed your branch is not mergeable "as is". Unfortunately I have no appropriate environment to test that, but I can already ask you the following changes:

* instead of adding 5 parameters to function pwg_db_connect, it would be better to have a single parameter $ssl_params, an array with 5 keys. And consequently the appropriate lines in local/config/database.inc.php:

```
$conf['db_ssl_params'] = array(
  'key' => null,
  'cert' => null,
  'ca' => null,
  'capath' => null,
  'cipher' => null,
);
```

* making sure the code will work fine even if there is no $conf['db_ssl_params'], because there is no need to alter this precious file in all Piwigo installations. so for example you can:

```
pwg_db_connect($conf['db_host'], $conf['db_user'],
               $conf['db_password'], $conf['db_base'],
               @$conf['db_ssl_params']);
```

and the function definition could be:

```
function pwg_db_connect($host, $user, $password, $database, $ssl_params=null)
```

Now one thing is maybe more concerning: the installation page needs to be ready for these SSL parameters, or else Piwigo would be just impossible to connect to a SSL remote database! And of course Piwigo install screen should not bother 99.99% users with these advanced parameters. So they must be hidden by default and visible on a action link like "show SSL connection options".

martinstefany:
Hi @plegall,
thanks a lot for feedback! I know it's not mergeable "as is", but I will work on it / improve it as you suggested.

cabotine:
Hi there,
is there a solution for a ssl connection to the myqsl database in Piwigo?
Best wishes

PR #562, opened 2016-12-05
Initial implementation of a "directory description file"

dirkhillbrecht:
Hello Piwigo maintainers,

I have written a small addition to the synchronisation process. I have a strictly upload-based workflow and it bothers me if I have to maintain some data of my photo archive only in the Piwigo database. Therefore I implemented a feature "directory description file": If a directory contains a file "00_index.txt", its first line is stored as "name" entry in the "piwigo_categories" table row for this directory. The third and all following lines are stored as "comment" field in the same table and row.

I am aware that the implementation is clumsy at best: I am not that big PHP expert and I am not sure whether this feature really belongs into Piwigo's core and not in a nice plugin. I am very, very, very open to any suggestions on how to clean up this code. Nevertheless, I've put this in the frame of a pull request hoping that this makes it easier for you to get an idea of what the heck I did after all and what the general idea of this code is.

I'd love to hear from you. Piwigo is a very very nice tool and with this extension, it becomes even more useful for workflows like mine.

Best regards,
Dirk Hillbrecht

### DIFF
--- a/admin/site_update.php
+++ b/admin/site_update.php
@@ -424,6 +424,48 @@ SELECT id_uppercat, MAX(rank)+1 AS next_rank
     $counts['new_categories'] = count($inserts);
   }
 
+  // Scan all directories for description file and insert its contents into name and comment field
+  // A directory description files contains a name and a comment for the directory
+  // which are inserted into the respective fields of the category table entires
+  // of Piwigo. The name is given in line 1. The comment is given in the lines 3
+  // and below. Line 2 must be empty. If only line 1 is given, the name is set and
+  // the comment is untouched.
+  // The index file is always named "00_index.txt".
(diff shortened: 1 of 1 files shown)


plegall:
Hi @dirkhillbrecht 

I think this feature is a good idea, but not the way it is implemented.

First, it should be a plugin, not a part of the core sync process.

Second, once the code is inside a plugin, you should first retrieve all physical categories (where dir is not null) and loop on it to search for a description file. And you should not use an SQL query to update each album, but fill an array and us a single mass_updates function.

If you need a model plugin, you can duplicate plugin Check Files Integrity http://piwigo.org/ext/extension_view.php?eid=844 and adapt it to your needs.

Such a plugin would be nice on piwigo.org/ext :-)

PR #570, opened 2016-12-16
Add RSS random feed

vhsantos:
Hello,

I was working in this request and attached is the patch.

Sorry, for the code...Im not a developer, I just copy and paste code from one place to other, but I think that is OK.. maybe a good developer can look and improve it.

to access the random feed:
http://site/feed.php?random

more info at http://piwigo.org/forum/viewtopic.php?id=26742

thanks and attentive.

### DIFF
--- a/feed.php
+++ b/feed.php
@@ -65,8 +65,21 @@ check_input_parameter('feed', $_GET, false, '/^[0-9a-z]{50}$/i');
 
 $feed_id= isset($_GET['feed']) ? $_GET['feed'] : '';
 $image_only=isset($_GET['image_only']);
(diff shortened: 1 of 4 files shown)


feed.php
```diff
@@ -65,8 +65,21 @@ function ts_to_iso8601($ts)
 
 $feed_id= isset($_GET['feed']) ? $_GET['feed'] : '';
 $image_only=isset($_GET['image_only']);
+$image_random=isset($_GET['random']);
```
rsalas: $image_random=isset($_GET['random'])?true:false;

and the "if" below is unnecesary.

Issue #220, opened 2015-07-25
[image processing] Generating thumbnails in Batch Manager not working on HTTPS ONLY setup


plegall:
**Reported by slima on 23 Mar 2013 14:23**

**Version:** 2.5.0

I have nginx server configured to serve content only via https.

Piwigo 2.5 Batch Manager dont generate thumbnails via &quot;generate multiple size images&quot; command.
No error in log, but I do research and it&#039;s HTTPS problem related. 

When I configure server to serve content via http generate working.

When I go to url:

https://myhost/g/i.php?/upload/2013/03/23/20130323121131-f50e5c66xx.jpg&amp;b=1364037169&amp;ajaxload=true 

I see :

{&quot;url&quot;:&quot;http:\/\/myhost:443\/g\/_data\/i\/upload\/2013\/03\/23\/20130323121131-f50e5c66-xx.jpg&quot;}

I think problem is &quot;http&quot; in url variable.

**Steps to reproduce:** 1) Go to https://galleryhost/admin.php?page=batch_manager
2) Select some image, and select in Action menu &quot;generate multiple size images&quot;
3) Select All
4) Click Apply Action button

Error: 7 photos can not be regenerated

[Piwigo Bugtracker #2879](http://piwigo.org/bugs/view.php?id=2879)

mossroy:
I face the same issue after a Piwigo upgrade.
I also see that some URLs point to http://mydomain/... instead of https://mydomain/...
I found this thread that seems related : http://piwigo.org/forum/viewtopic.php?id=25782 but I did not manage to make it work so far

plegall:
I will soon work on implementing HTTPS with Nginx. I will have to fix that problem, for sure!

mossroy:
Great!
In my case, it's an Apache reverse-proxy (on another server, which handles SSL encryption), but I suppose the fix would be the same for both HTTP servers.
I'm using Piwigo 2.7.4 on Linux (Debian Wheezy)
PHP: 5.4.41-0+deb7u1
MySQL: 5.5.44-0+deb7u1
Bibliothèque graphique: ImageMagick 6.7.7-10

EvKoh:
+1

arigit:
Same problem. NGINX + HTTPs, and started to have problems with Piwigo, thumbnails that fail to get generated at random, it's weird. Clearly related to https. My site supports both HTTP and HTTPs access.

ReSearchITEng:
facing the same (with https).
[...]

PR #597, opened 2016-12-30
Better support for front HTTP proxy

byo:
Signed-off-by: Bartłomiej Święcki <byo.spoon@gmail.com>

### DIFF
--- a/include/config_default.inc.php
+++ b/include/config_default.inc.php
@@ -590,6 +590,15 @@ $conf['picture_url_style'] = 'id';
 // tags is not unique, all tags with the same url representation will be shown
 $conf['tag_url_style'] = 'id-tag';
 
+// allow_http_proxy_servers : this option configures front HTTP proxy server
+// support. If set to false, piwigo assumes it's not running behind http proxy.
+// If set to true, piwigo will assume it is running behind one and thus it will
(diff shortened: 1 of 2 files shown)


byo:
Hi @modus75, can you explain a bit more your concerns?

Exactly because you can manually generate X-Forwarded-... headers when forming http requests I've made this feature disabled by default. And even if you want to have this enabled, you can choose a list of IP addresses which are genuine reverse http proxies, only those will be able to provide X-Forwarded-... header taken into consideration.

I see those headers easily accepted by many apps written in RoR or Java (I was able to make them work without any tweaks or with little config adjustments only, all based on X-Forwarded-... headers).

If you really think that putting this into local configuration file is a better idea then I think the X-Forwarded-Host header should also not be taken into account, which is the case in [current vanilla master code](https://github.com/Piwigo/Piwigo/blob/c789347c514e89b32c9ceefa0a9c9ac819cbc677/include/functions_url.inc.php#L67). My motivation for upsteaming this patch was that relying on X-Forwarded-Host only is not enough to handle all reverse proxies (it works for basic cases only).

modus75:
You're right the forwarded should not be there... I'll have to check the history and see...

hg8496:
I think it's a clean solution. It would also fix #220 which annoys me for quite some time.

Issue #250, opened 2015-07-25
[synchronization] Sync fail with files containing ' (apostrophe) in filename


plegall:
**Reported by JanisV on 23 Jan 2014 11:06**

**Version:** 2.6.0

Sync any folder with file &quot;Kamen&#039; at Nameless.JPG&quot; and $conf[&#039;sync_chars_regex&#039;] = &#039;/.*/&#039;; fail with:
Warning:  [mysql error 1064] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near &#039;at Nameless.JPG&#039;,&#039;Kamen&#039; at Nameless&#039;,&#039;2014-01-22 23:45:16&#039;,&#039;./galleries/Photo/&#039; at line 11555

**Additional information:** piwigo/admin/site_update.php:512 use filename without pwg_db_real_escape_string, like in piwigo/admin/include/functions_upload.inc.php:354

[Piwigo Bugtracker #3030](http://piwigo.org/bugs/view.php?id=3030)

plegall:
**Comment posted by JanisV on 23 Jan 2014 11:13**

Related issue #2971

plegall:
**Comment posted by bryanw on 25 Sep 2014 20:47**

It is crashing because you modified the regex from the default.  Set it back to the default value of:  /^[a-zA-Z0-9-_.]+$/

That is, the system only accepts a range of the English a-z and A-Z, the numbers 0-9, the underscore, the single dash, and the dot/period. No other characters are allowed. No spaces. No accents like é. No slashes (forward or back). No symbols. Using them can cause the app to crash or the database to become corrupt or other bad things to happen.

Unfortunately bad handling of input is pervasive through the Piwigo code. 

Apostrophe, accented characters, slashes, and many other characters cause serious problems and the dev team does not want to devote the resources to rewrite large segments of the code. Remember it is a small team of volunteers. 

In addition to file names it also impacts email addresses, user comments, metadata, and more. 

Since the dev team is small I&#039;ve looked into fixing it and posting a patch (I&#039;m a programmer by day) but the extent of the change is huge. It impacts almost every part of the program. Nearly everything in the system touches the database, and nearly all of the database code improperly handles queries. It is not a simple bug fix, but a near-complete rewrite of the entire database interface.

[...]

Issue #544, opened 2016-10-28
SQL Injection 'may' be possible through filenames


vrdriver:
Hello, I thought I'd let you know of a problem I just encountered.

While my suggestion is a long shot, I had the following set of errors while attempting to do a "Database synchronisation with files: ./galleries/"

Warnings from the program. 

```
[./galleries/all_low/29_Tel_Beit_She'an_(Decapolis)] PWG-UPDATE-1 (wrong filename)
[./galleries/all_low/51_Hezekiah's_Tunnel] PWG-UPDATE-1 (wrong filename)
[./galleries/all_low/73_Pre-Dinner_stop_(Al_Hasa,_Jordan)] PWG-UPDATE-1 (wrong filename)
[./galleries/all_low/Items_From_31_Qasr_El_Yahud_(Jordan_Baptism_Site)] PWG-UPDATE-1 (wrong filename)
[./galleries/all_low/Items_From_31_Qasr_El_Yahud_(Jordan_Baptism_Site)/31_Qasr_El_Yahud_(Jordan_Baptism_Site)] PWG-UPDATE-1 (wrong filename)
```

The actual SQL Error:

`Warning:  [mysql error 1064] You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'an_(Decapolis)/2016_09_18_-_Day_06/2016-09-18_20160918_055416.jpg',NULL,'286','1' at line 1084`

```

INSERT  INTO piwigo_images
  (id,file,name,date_available,path,representative_ext,storage_category_id,added_by)
  VALUES('713','2016-09-13_0K3A0165.jpg','2016-09-13 0K3A0165','2016-10-28 15:12:51','./galleries/all_low/00_Transit/2016_09_13_-_Day_01/2016-09-13_0K3A0165.jpg',NULL,'331','1')
  , ('714','2016-09-13_0K3A0167.jpg','2016-09-13 0K3A0167','2016-10-28 15:12:51','./galleries/all_low/00_Transit/2016_09_13_-_Day_01/2016-09-13_0K3A0167.jpg',NULL,'331','1')
  , ('715','2016-09-13_20160913_145837.jpg','2016-09-13 20160913 145837','2016-10-28 15:12:51','./galleries/all_low/00_Transit/2016_09_13_-_Day_01/2016-09-13_20160913_145837.jpg',NULL,'331','1')
[...]

PR #598, opened 2017-01-02
mysqli: escape raw string values

daniele-athome:
I've implemented use of `pwg_db_real_escape_string` in cases where raw strings were injected into SQL code. This will prevent SQL injection and SQL syntax errors when having quotes in filenames (if allowed by the allowed filenames regex).

I've tested it functionally and it works, but I couldn't test it in the system, e.g. I don't know if some input values are subjected to `addslashes` before being used in SQL. It certainly works for sync (mass_inserts). Anyway I believe that any `addslashes` use should be removed when dealing with the database and we should always use the dbms escape tools.

Ref: #250, #544

Signed-off-by: Daniele Ricci <daniele.athome@gmail.com>

### DIFF
--- a/include/dblayer/functions_mysqli.inc.php
+++ b/include/dblayer/functions_mysqli.inc.php
@@ -311,7 +311,7 @@ UPDATE '.$tablename.'
 
         if (isset($data[$key]) and $data[$key] != '')
         {
-          $query.= $separator.$key.' = \''.$data[$key].'\'';
+          $query.= $separator.$key.' = \''.pwg_db_real_escape_string($data[$key]).'\'';
         }
         else
         {
@@ -338,7 +338,7 @@ UPDATE '.$tablename.'
           }
           if (isset($data[$key]))
           {
-            $query.= $key.' = \''.$data[$key].'\'';
+            $query.= $key.' = \''.pwg_db_real_escape_string($data[$key]).'\'';
           }
           else
           {
@@ -453,7 +453,7 @@ UPDATE '.$tablename.'
 
     if (isset($value) and $value !== '')
     {
-      $query.= $separator.$key.' = \''.$value.'\'';
+      $query.= $separator.$key.' = \''.pwg_db_real_escape_string($value).'\'';
     }
     else
     {
@@ -481,7 +481,7 @@ UPDATE '.$tablename.'
       }
       if (isset($value))
       {
-        $query.= $key.' = \''.$value.'\'';
+        $query.= $key.' = \''.pwg_db_real_escape_string($value).'\'';
       }
       else
       {
@@ -556,7 +556,7 @@ INSERT '.$ignore.' INTO '.$table_name.'
         }
         else
         {
-          $query .= "'".$insert[$dbfield]."'";
+          $query .= "'".pwg_db_real_escape_string($insert[$dbfield])."'";
         }
       }
       $query .= ')';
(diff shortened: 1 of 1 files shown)


plepe:
This branch looks good, and it works in my environment. How are changes that it will be merged soon?

samwilson:
This PR contains lots of commits that are not part of it, and also is not able to merge. It should be rebased and have only topical commits (probably just one).

(Not that I'm a maintainer or anything, but that's what I'd say if I were reviewing this.)

daniele-athome:
My only commit is in fact 26aaea73e5122a29d9ef02b2dead0834619e3bc2, probably the upstream branch was rebased or history rewritten, therefore the commit mess. I'll create a new PR - or try to fix this one.

daniele-athome:
I've rebased my modifications onto the current master.

plepe:
With this commit active, synchronization of files which have a ' in their name does work, but on the other hand if you edit albums and add ' to a value (name, description), they will be saved escaped ("foo'bar" will turn into "foo\'bar", save again and it will turn into "foo\\\'bar").

So, I used the commit while synchronizing and disabled it again afterwards, as everything seems to work as expected.

daniele-athome:
> 
> 
> With this commit active, synchronization of files which have a ' in their name does work, but on the other hand if you edit albums and add ' to a value (name, description), they will be saved escaped ("foo'bar" will turn into "foo'bar", save again and it will turn into "foo\'bar").
> 
> So, I used the commit while synchronizing and disabled it again afterwards, as everything seems to work as expected.

I didn't try it on albums, thanks for the feedback. I'll test it out some more and check for more impacts. I thought it would have been too easy eheh

Cygon:
Suggestion: **don't** escape string values.

Use prepared statements with separate parameters, i.e. https://stackoverflow.com/questions/60174/how-can-i-prevent-sql-injection-in-php/60496#60496

No maze of escape/unescape methods and rules, unsupported mysql_ PHP extensions and completely safety from SQL injection attacks. Less maintenance / changes to the original code, too.

obones:
> > 
> 
> I didn't try it on albums, thanks for the feedback. I'll test it out some more and check for more impacts. I thought it would have been too easy eheh

I believe this is because your patch changes the generic insert/update query handlers but most code using those functions is already calling `pwg_db_real_escape_string` on values before passing them to the functions you patched.

So, if I'm correct, there are multiple options:

1. Remove all calls to `pwg_db_real_escape_string` and apply your patch
2. Remove all calls to `pwg_db_real_escape_string` and move to prepared statements (if they are available in all possible databases)
3. Don't apply your patch but rather track places where `pwg_db_real_escape_string` should have been called and add the missing call.

Personally, I went for option 3 around line 534 in `site_update.php` which now looks like this:

    $filename = pwg_db_real_escape_string($filename);
    $insert = array(
      'id'             => $next_element_id++,
      'file'           => $filename,
      'name'           => get_name_from_file($filename),
      'date_available' => CURRENT_DATE,
      'path'           => pwg_db_real_escape_string($path),
      'representative_ext'  => $fs[$path]['representative_ext'],
      'storage_category_id' => $db_fulldirs[$dirname],
      'added_by'       => $user['id'],
      );


Sadly, I'm quite sure I have missed other locations where this issue is present but it fixed Issue #250

yogeshvar (approved):

plegall:
I agree with @obones : modifying the generic insert/update query handlers is not the right solution. We should call `pwg_db_real_escape_string` in the right places.

PR #614, opened 2017-02-03
Include the current user's id in the API session info call

abrenoch:
I noticed there isn't a great way to get the current user's id (that I noticed anyway), so I thought this would be a good place to include it!


abrenoch:
Oops, looks like my other pull request got included - will close and try again!

PR #616, opened 2017-02-06
Include the current user's id in the session info

abrenoch:
I couldn't find a straightforward way of getting the current user's id, so this seemed like the most reasonable place to propose the change. This would be useful when using the api to call other methods that might require a user id (such as the user collections extension)

### DIFF
--- a/include/ws_functions/pwg.php
+++ b/include/ws_functions/pwg.php
@@ -317,6 +317,7 @@ function ws_session_getStatus($params, &$service)
   global $user, $conf;
 
   $res['username'] = is_a_guest() ? 'guest' : stripslashes($user['username']);
+  $res['user_id'] = is_a_guest() ? 'guest' : $user['id'];
   foreach ( array('status', 'theme', 'language') as $k )
   {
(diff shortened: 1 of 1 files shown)


flop25:
Seems a good idea.
However the guest user does have an id so you can answer back its id instead of 'guest'

plegall:
@abrenoch why do you need the user id? or, asked another way, why would Piwigo expose this internal user id to non administrators?

abrenoch:
@plegall I was working on adding support for the user collections plugin to the Kodi add-on I manage, and had noticed if you invoke the `pwg.collections.getList` as an admin with no `user_id` supplied, then the plugin returns the collections of all users. Granted this is indeed a flaw with the user collections extension it seems; I had figured having access to that piece of information may be helpful in supporting additional extensions in the future - but I can understand where you may not want to include it!

plegall:
For me it's a problem with `pwg.collections.getList` : even as an admin, it should return the list of collections for the current user. Only if the user_id is provided it should list the collections of the requested users.

PR #715, opened 2017-06-23
Solved problem with deprecated warning

rsalas:
Solved problem with deprecated warning.
Update for PHP 7.

### DIFF
--- a/include/ws_core.inc.php
+++ b/include/ws_core.inc.php
@@ -401,7 +401,10 @@ Request format: ".@$this->_requestFormat." Response format: ".@$this->_responseF
 
   static function isPost()
   {
-    return isset($HTTP_RAW_POST_DATA) or !empty($_POST);
+    if (filter_input(INPUT_SERVER, 'REQUEST_METHOD') === 'POST') {
+      return true;
+    }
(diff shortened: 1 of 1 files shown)


include/ws_core.inc.php
```diff
@@ -401,7 +401,10 @@ function getMethodOptions($methodName)
 
   static function isPost()
   {
-    return isset($HTTP_RAW_POST_DATA) or !empty($_POST);
```
samwilson: I think this can just be `return $_SERVER['REQUEST_METHOD'] === 'POST';`

That still uses a global variable, but at least it's not deprecated.

Note that filter_input() doesn't work with INPUT_SERVER or INPUT_ENV when you use FASTCGI:  https://stackoverflow.com/a/36205923/99667

samwilson: I think this can just be return $_SERVER['REQUEST_METHOD'] === 'POST';

That still uses a global variable, but at least it's not deprecated.

Note that filter_input() doesn't work with INPUT_SERVER or INPUT_ENV when you use FASTCGI: https://stackoverflow.com/a/36205923/99667

Issue #731, opened 2017-07-07
dashboard, missing number of photos


plegall:
If I have 1945 photos, the template variable NB_PHOTOS is 1,945 and it's not >1, so it's not displayed. We should distinguish the number of photos and the number for display.

PR #754, opened 2017-09-03
Fixing issue #731

MaximeBOURMAUD:
if NB_PHOTOS is upper than 1000 display it.


plegall:
We don't want to show the number of photos if there is only one, but we want to show the number of photos if you have between 1k and 2k photos. Fake the number before sending it to the template and you'll see the test in the template is inappropriate.

PR #770, opened 2017-09-15
Feature#765 now it's possible to login with your email address

### LINKED ISSUE #765: ability to log in with email address (instead of username)
Several users have reported authentication failure. Piwigo was working as expected, the users were simply using their email address as username (as many web applications do).

It would be more convenient if Piwigo could allow/understand when a visitors tries to authenticate with its email address (that he never forgets, as opposed to the Piwigo username).

For now, the problem I see is that you can use an email address as username. Imagine 2 users 👍 

* username "alfonso" with email address "alfonso@protonmail.com"
* username "alfonso@protonmail.com" with email address "very@funny.com"

Then what happens when you submit "alfonso@protonmail.com" on identification.php ? I would say, we match in priority the username (which does not follow an email address pattern 99.999% of the time) and if nothing matches, try among email address.

Reminder: email addresses must be unique in Piwigo.

maximebrmd:
Fixes #765

### DIFF
--- a/include/functions_user.inc.php
+++ b/include/functions_user.inc.php
@@ -1112,22 +1112,23 @@ function pwg_login($success, $username, $password, $remember_me)
 
   global $conf;
 
-  $user_found = 'false';
+  $user_found = false;
   // retrieving the encrypted password of the login submitted
   $query = '
 SELECT '.$conf['user_fields']['id'].' AS id,
        '.$conf['user_fields']['password'].' AS password
   FROM '.USERS_TABLE.'
   WHERE '.$conf['user_fields']['username'].' = \''.pwg_db_real_escape_string($username).'\'
-  ;';
+;';
 
   $row = pwg_db_fetch_assoc(pwg_query($query));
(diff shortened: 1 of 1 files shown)


plegall:
The changes introduced in https://github.com/Piwigo/Piwigo/pull/770/commits/2999b6fee941d9cf82b8c48de24c427e64d838c6 are not enough and may lead to an error. Take the case of user "alfonso@protonmail.com" (yes, a username can look like an email address) where another user "alfonso" has email address "alfonso@protonmail.com" => 2 lines from table users will be returned by the SQL query but we only care about the first one in PHP... You can't know which one will be returned first. A legitimate email/password login may be refused :-/

plegall:
Please do not duplicate the {log_user, trigger_notify, return true} block, do it only once to avoid any problem in the future (if someone adds an instruction in the first block, (s)he will certainly forgot to add it in the second block, so let's have only one block.

plegall:
maybe "$bool" is not the most appropriate variable name :-) Something like "$user_found" would be much more self-explanatory :-)

Issue #686, opened 2017-05-22
admin dashboard, first photo thumbnail


plegall:
in the tooltip of the "first photo added" block, it would be fun to display the thumbnail of this first photo !

PR #776, opened 2017-09-19
New feature, display first photo thumbnail

maximebrmd:
#686 New feature


plegall:
There is already another SQL query. One (when modified) should be enough (the less SQL queries the better)

plegall:
After a very long thought, I've decided to change the way we provide the age of install, see #1707 

This pull-request is no longer relevant :-/

PR #784, opened 2017-09-27
Issue/664 remember the list of active extensions

MaximeBOURMAUD:


plegall:
@MatthieuLP this pull-request no longer matches the new plugin manager. How could we imagine having this feature with the new plugin manager?

PR #785, opened 2017-09-27
issue #544 Check escape string

maximebrmd:


plegall:
Same as #598 : modifying the generic insert/update query handlers is not the right solution. We should call `pwg_db_real_escape_string` in the right places.

Issue #737, opened 2017-07-21
Embed Open Graph protocol


You85:
"Officially, the Opengraph has no impact on SEO, but social networks are an excellent source of traffic, and if your content is more visible on Facebook they will be much more shared, Get a much better click-through rate and therefore an increase in traffic and conversions, a relevant indicator that Google can take into account, so the impact on SEO is rather indirect."

http://ogp.me/

nekohayo:
Actually this is a duplicate of issue #239, close it as such?

PR #786, opened 2017-09-27
Add twitter cards and Open Graph on pages

tcitworld:
Adds Open Graph Protocol tags and Twitter Cards support to the main template. If we are on a picture page, we also have it's preview.

Fix https://github.com/Piwigo/Piwigo/issues/737

### DIFF
--- a/themes/default/template/header.tpl
+++ b/themes/default/template/header.tpl
@@ -18,6 +18,26 @@
 {/if}
 {/if}
 
+{if isset($COMMENT_IMG)}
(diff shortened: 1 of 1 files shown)


TallGuysFreeADMIN:
I think this is better as an extension. Many people, including myself, choose Piwigo for simplicity. 

Piwigo galleries might be:
Private, locally hosted, or on an intranet. 

If anyone is using HTML in their page descriptions, this isn't going to work nicely. 

Having the twitter site as @piwigo is very unacceptable.

Twitter Cards is already an extension.

flop25:
of course we won't implement twitter or any website specific html
only opengraph tags could be added
as it this commit will never be accepted

PR #801, opened 2017-11-14
Sanitize plugin ID for maintenance class name

ejegg:
In case a plugin has e.g. a hyphen in its ID


plegall:
Hi @ejegg 

(OK, it took many years for us to reply, but I hope you're still there)

Do you have a use case to explain this change? Isn't it dangerous to live change the name of the class, instead of having it right from scratch?

Issue #778, opened 2017-09-20
Retrieve untagged images via the API


abrenoch:
I am working on a project and would really like to be able to retrieve images that are not tagged/keyworded (like in the batch manager) via the API. I'm willing to do the work and make a pull request, but I wanted to establish the most favorable way of accomplishing this.

I'm inclined to think the most logical way to request this info would be to use the `pwg.tags.getImages` method and provide `null` for the `tag_id` parameter - which is not currently supported.

@plegall would you have any input on this? Thanks!

abrenoch:
Any input on this guys? Would love to make this work!

PR #808, opened 2017-11-26
Modified pwg.tags.getImages to allow returning images that are not tagged

abrenoch:
https://github.com/Piwigo/Piwigo/issues/778

Just let me know if there would be a more preferable way of achieving this!

### DIFF
--- a/include/ws_functions/pwg.tags.php
+++ b/include/ws_functions/pwg.tags.php
@@ -93,57 +93,79 @@ function ws_tags_getAdminList($params, &$service)
  */
 function ws_tags_getImages($params, &$service)
 {
-  // first build all the tag_ids we are interested in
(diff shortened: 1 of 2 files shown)


abrenoch:
It appears some recent changes may have broken this - `WS_TYPE_INT|WS_TYPE_POSITIVE` used to work with `null` but that does not appear to be the case anymore. Will revise.

abrenoch:
@plegall what would be a preferable alternative to providing `null` to the `tag_id`? It seems to make that work we would either need a different constant (something like `WS_TYPE_NULL`), or I can revise the API call to have a new parameter to fetch the untagged images; something like `untagged` that defaults to `false`.

Seems like adding the new parameter might be the more desirable choice!

PR #810, opened 2017-11-28
API call to add/remove favorites

abrenoch:
Implements `pwg.users.addFavorite` and `pwg.users.removeFavorite`


include/ws_functions/pwg.users.php
```diff
@@ -621,4 +621,53 @@ function ws_users_setInfo($params, &$service)
...
+function ws_addFavorite($params, &$service)
+{
+  global $user;
+  if (is_a_guest())
+  {
+    return new PwgError(403, 'User must be logged in.');
+  }
+
+  $query = '
+INSERT INTO '.FAVORITES_TABLE.'
```
samwilson: Should this be 'insert ignore'? What happens if the favorite is already there?

abrenoch: Probably should. I can make the adjustments, however I'm not exactly confident any of these PRs will ever make it to the project at this point.

samwilson: Yeah, things can be slow sometimes I guess. But it'll never be merged if it's not correct. :)

PR #816, opened 2017-12-09
Add Docker support

J0WI:
This adds [Docker](https://www.docker.com/) support to Piwigo.

Sadly I failed to add a database setup to `docker-entrypoint.sh`, because the installation process requires to write some values into the `piwigo_config` table and a admin account. If you only create the `local/config/database.inc.php` you will end up in a broken Piwigo instance.

It was also not clear to me if Piwigo supports other DB's than MySQL, so the container currently doesn't support PostgreSQL or SQLite. It would be trivial to add them too.
ImageMagick is installed since it's recommend by Piwigo, but it would also work fine with PHP GD.

It would be great if you could do the steps to get into Dockers standard library aka "Official Images".

There are some administrative steps to fulfill:
https://github.com/docker-library/official-images#contributing-to-the-standard-library

This will enhance maintenance and trust in you Docker image.


zicklag:
Hi @J0WI, I like your Docker image and I would love to have an official Piwigo container as I was just writing my own for Piwigo. I did have a couple of thoughts on your image that you might consider.

### Copying Piwigo from the source repository

Instead of downloading Piwigo from the website, you could copy the Piwigo source into the container from the source repository. You would probably want and/or need to move the Dockerfile to the root directory of the repository where you could easily copy the whole thing into the container.

With this model you can build a docker container for any version of Piwigo just by checking out a different branch/tag.

### More Specific Docker Volumes

In your container you persist the whole Piwigo installation in a volume. This does allow you to use Piwigo's builtin updates from the admin interface, but IMHO I think it is generally better to update your container image for software updates as opposed to storing the software installation in a volume. I *do* believe that you should create volumes for local configuration, the galleries, the uploads, the themes, the plugins, and the thumbnail cache ( `_data/i` ). This separates the application data from the application itself, which I think is one of the great advantages of Docker.

These are just my suggestions from my own experience with Docker. Tell me what you think. :smiley:

J0WI:
> Copying Piwigo from the source repository

My goal is that this image can be published in the docker store, so it's required to support automated builds. You can still use something like `--volume .:/var/www/html` if you need a custom branch.

> More Specific Docker Volumes

You can still do this with something like `--volume myimages:/var/www/html/_data/i`.
I like this way because it's the most flexible. You can even use a PHP FPM variant of the image without any big changes. I don't know why other "official" docker images are doing it the same way, but I'm sure there are other good reasons.

zicklag:
OK, that's cool. I love Docker and am always interested to see the different ways people write their containers. I'm still experimenting and finding new techniques to write them. :wink:

J0WI:
The best would be to have a dedicated repo for the Docker stuff.

J0WI:
@plegall is there any chance that you bring this to Dockerhub?

keachi:
:+1:

Phlogi:
https://hub.docker.com/r/linuxserver/piwigo

PR #866, opened 2018-05-07
Include pwg_token in user list POST request (Fixes #748)

### LINKED ISSUE #748: Unable to view users - 403 on user_list_backend.php
Hi.  I just setup a no piwigo gallery and am trying to create a new user.  When browsing the Users, Manage, page, though, I get this error message:

`DataTables warning: table id=userList - Ajax error. For more information about this error, please see http://datatables.net/tn/7`

Tried creating a user anyway and that part actually worked (I can see the new user record in the database and login as that user), but I still get that error every time I try to view the Users, Manage page, and thus I'm unable to view or manage the user.

I checked the server logs and found that I"m getting a 403 when trying to POST to admin/user_list_backend.php:

`[30/Aug/2017:03:04:29 -0500] "POST /photos/admin/user_list_backend.php HTTP/1.1" 403 243 "https://www.website.net/photos/admin.php?page=user_list"`

I loaded Firebug in Firefox and can see the 403, but I can't find any extra insight into what's causing it.  Pasting a screenshot that shows the headers for both admin.php?page=user_list as well as the follow-up request to admin/user_list_backend.php in the hope it may show you something useful.

image

Any idea what may be causing this?  or what I can do to troubleshoot further?  This is my first piwigo instance, so I don't have have another instance to compare against.

For reference, this is running under apache 2.4.27 w/ PHP 7.0.21 on a Linux host.  Happy to provide more details if it will help.

Thanks.

dadap:
The addition of the POST data uses the newer DataTables 1.10 API. The existing code seems to use a deprecated API, even though Piwigo uses DataTables 1.10, and I couldn't figure out how to get the POST data correctly added using the older API.

### DIFF
--- a/admin/themes/default/template/user_list.tpl
+++ b/admin/themes/default/template/user_list.tpl
@@ -615,12 +615,12 @@ jQuery(document).on('click', '.close-user-details',  function(e) {
     processing: true,
     serverSide: true,
 		serverMethod: "POST",
-    "ajax": {
-        "url" : "admin/user_list_backend.php",
-         "type" : "POST",
-         "data" : {
-            "pwg_token" : pwg_token
-         }
+    ajax: {
+        url : "admin/user_list_backend.php",
+        type : "POST",
+        data : {
+            pwg_token : pwg_token
+        }
     },
 		pagingType: "simple",
     language: {
--- a/admin/themes/default/template/user_list.tpl
+++ b/admin/themes/default/template/user_list.tpl
@@ -615,7 +615,13 @@ jQuery(document).on('click', '.close-user-details',  function(e) {
     processing: true,
     serverSide: true,
 		serverMethod: "POST",
-    ajaxSource: "admin/user_list_backend.php",
+    "ajax": {
+        "url" : "admin/user_list_backend.php",
+         "type" : "POST",
+         "data" : {
+            "pwg_token" : pwg_token
+         }
+    },
 		pagingType: "simple",
     language: {
       processing: "{/literal}{'Loading...'|translate|escape:'javascript'}{literal}",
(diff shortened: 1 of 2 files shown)


dadap:
I added a third commit to this PR (761b27986) after realizing that a235a2e22 included quotes around the object key names that weren't necessary. (I was treating the "ajax" object like JSON rather than JavaScript.) Feel free to squash those commits together if you end up merging this PR.

plegall:
I don't understand why pwg_token would be necessary to `user_list_backend.php` but it can't hurt :-)

plegall:
Unfortunately, the change from "ajaxSource" to "ajax" in datatables has completely changed the request to admin/user_list_backend.php. Pagination, search, sorting no longer work on Piwigo 2.9.4. I should have made more tests. The user list was apparently working.

dadap:
That’s an unfortunate side effect. Can you still add the pwg_token without switching the datatables API? The only reason I did that was because I was unfamiliar with datatables and from the docs I couldn’t figure out how to add POST data without switching to the newer API.

As for the user list, it was most definitely not working for me or any of the others who reported the bug. I suspect some piece of server configuration might be responsible though I didn’t investigate to determine which.

plegall:
If the pwg_token (I still don't understand why it's needed here) is OK via GET instead of POST, this would work:

```
ajaxSource: "admin/user_list_backend.php?pwg_token="+pwg_token,
```

dadap:
I can try testing that later today. Wouldn’t that be less secure, though? Revealing the token in the GET string seems risky.

plegall:
Absolutely not a security problem :-)

Yes please test that solution instead of `ajax : ...`: if it works I will change for Piwigo 2.9.5

plegall:
@dadap did you get the opportunity to test?

dadap:
Sorry, not yet. I’ll try to do this soon.

dadap:
Your suggestion does work, but only when the HTTP method is also changed from POST to GET. With GET instead, the pwg_token data doesn't need to be provided, as you expected. See PR #903 for a new change that's also working for me.

Issue #867, opened 2018-05-09
increase PHP version required (to 7.0)


plegall:
Piwigo 2.9.3 is still marked as compatible with PHP 5.2. Oldest officially maintained branch is PHP 5.6. Wordpress.org statistics https://wordpress.org/about/stats/ currently says that 5.4+ is 89% of users and 5.5+ is 76%.

I think increasing the required version to 5.4 is "nearly painless".

plegall:
Interesting study on PHP versions https://seld.be/notes/php-versions-stats-2018-1-edition

plegall:
Considering the following facts:

- Piwigo 2.10 is not for "soon" (not this summer I'm afraid)
- compatibility with PHP 7.2 is becoming problematic (Ubuntu 18.04 comes with PHP 7.2)
- compatibility with PHP 7.2 breaks compatibility with PHP 5.2 (requires PHP 5.3)

I'm trying to merge the current pull-request  #834  for branch 2.9, which will require PHP 5.3.

plegall:
See b1cb4e48c03f4ce62b3de66c3337d8b9ef656477

J0celyn:
Note that PHP 5.6 and PHP 7.0 are now no longer maintainted.
Sources: http://php.net/eol.php and http://php.net/supported-versions.php

plegall:
@J0celyn that's right... in theory. In practice however, there are still a vast majority of users running a PHP version older than 7.1. See https://wordpress.org/about/stats/ : 76,2% of users are running an "officially unsupported" version of PHP. We have to deal with that :-)

J0celyn:
@plegall 
My previous message is not theory, it is factual information given on http://php.net about which PHP versions they still support as well as an update to your first post stating "Oldest officially maintained branch is PHP 5.6".

After looking at the Wordpress usage stats I understand your motivation to maintain PHP 5.x compatibility in Piwigo and providing support for users running an old version of PHP.

[Wordpress requirements](https://wordpress.org/support/article/requirements/) page clearly warns about the risks when running old versions of PHP/MySQL:
> Note: If you are in a legacy environment where you only have older PHP or MySQL versions, WordPress also works with PHP 5.2.4+ and MySQL 5.0+, but these versions have reached official End Of Life and as such **may expose your site to security vulnerabilities**.

plegall:
@J0celyn OK, we will add this warning

plegall:
[...]

PR #875, opened 2018-06-05
Use entire language code.

xppppp:
HTTP_ACCEPT_LANGUAGE="ko" would match "kok_IN" instead of correctly matching "ko_KR"

### DIFF
--- a/include/functions_user.inc.php
+++ b/include/functions_user.inc.php
@@ -814,7 +814,7 @@ function get_browser_language()
   $browser_language = substr(@$_SERVER["HTTP_ACCEPT_LANGUAGE"], 0, 2);
   foreach (get_languages() as $language_code => $language_name)
   {
-    if (substr($language_code, 0, 2) == $browser_language)
+    if (explode("_", $language_code)[0] == $browser_language)
     {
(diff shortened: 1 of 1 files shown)


plegall:
Function array dereferencing has been added in PHP 5.4 and Piwigo is currently (version 2.9) not requiring such a version but 5.2. We plan to increase the requirement to PHP 5.4 #867 for Piwigo 2.10 so I think we can merge this pull-request :-)

xppppp:
I just noticed that this PR is still out there.  In any case, language codes aren't always two characters, per RFC3066 (https://tools.ietf.org/html/rfc3066 - which references https://www.loc.gov/standards/iso639-2/php/code_list.php)

samwilson:
Here's another regex for pulling language info from there: https://github.com/Krinkle/intuition/blob/236b5b134e045a109f85901ce4f50e70894db9be/src/Util.php#L159 (that Intuition library is pretty good at handling languages well; I've used it in a few projects).

plegall:
I close this pull-request since #1132 does a more in-depth job (and was just merged)

PR #894, opened 2018-07-12
Fix syntax error

akimitsu-bm:
> Parse error: syntax error, unexpected ')' in


plegall:
You're the boss. I was testing Piwigo before releasing 2.9.4 and you have found a blocking bug (+ code fixing). Thank you :-)

Issue #817, opened 2017-12-09
Add Travis CI


J0WI:
How about using Travis CI?

https://docs.travis-ci.com/user/getting-started/

plegall:
A first step would be to have automated tests ;-)

@MaximeBOURMAUD has started working on a script which installs + creates an album + adds photo. This could be a first step to automation. But we're far from Travis I think!

samwilson:
Even without tests, adding CI could help prevent things like #894, e.g. even if it did nothing other than run `jakub-onderka/php-parallel-lint`. It could also run phpcs.

samwilson:
I've had a crack at this in #895.

PR #895, opened 2018-07-13
Add Github Actions config for syntax checking

samwilson:
This adds continuous-integration configuration for checking PHP syntax.

The CI runs can be seen at https://github.com/samwilson/Piwigo/actions/workflows/ci.yml

Refs: #817


samwilson:
Lots of these tests fail at the moment :-) I temporarily added my fork to Travis https://travis-ci.org/samwilson/Piwigo and you can see an example of the results at https://travis-ci.org/samwilson/Piwigo/jobs/403372472

flop25:
99.9% of 'Line indented incorrectly'
thanks for your contributions

samwilson:
I'm not sure if you meant it, but that comes across as sarcastic. I'm actually trying to help here!

I'm not suggesting that the phpcs ruleset is complete or correct, and of course it needs to be fixed up. The idea here is that automatic checks be done. The details of those checks need to be figured out. :-)

flop25:
I'm not suggesting anything neither

samwilson:
> I'm not suggesting anything neither
Okay, sorry, I took your comment wrongly! :-) Is easily done in these text worlds.

Anyway, I've removed the indenting rules. There are still lots of errors, but they can all (I think) be fixed automatically.

There are lots of files in `language/` with their executable bits set.

samwilson:
I've remove all but the syntax checking from this PR, to keep it simpler. The build is now passing:
https://travis-ci.org/samwilson/Piwigo/builds/403788915

The changes to xmlrpc_encoder.php are to wrap the  xmlrpc_encode() function in a conditional, but I didn't re-indent the actual function just to keep the diff cleaner.

plegall:
Adding Continuous Integration with automated check is great. I'll review it next week. Thank you @samwilson

samwilson:
I've brought this up to date, in case it's still of interest. See example results at https://travis-ci.org/samwilson/Piwigo/builds/571741891

samwilson:
I've brought this up to date and switched it to use Github Actions. Results are at https://github.com/samwilson/Piwigo/actions/runs/127365045

plegall:
Hi @samwilson

I'm reviewing this pull-request. I don't know yet how to deal with it. I think it would be interesting indeed to have an automatic check on syntax and maybe on coding style (which would already be complex to configure I guess).

samwilson:
I've added PHP 8.0 and 8.1 to this patch now (and removed the comment about only testing against supported PHP versions, because it's good to keep testing against all working versions). You can see the current results at https://github.com/samwilson/Piwigo/actions/runs/2729396577

I think having at least a basic syntax-check would be a good thing (to avoid issues such as  #894), and as you say it could eventually be extended to also include coding style (I started some work on a phpcs standard at https://github.com/samwilson/piwigo-coding-standards ).

Let me know if there's more info that should be added to this patch.

Issue #902, opened 2018-07-24
[2.9.4] users list no longer paginated or filtered


plegall:
On Piwigo 2.9.4, compared to 2.9.3, the users list can no longer be filtered and is not paginated. It always shows the complete list of users. If you have 15 users that's not a big problem. If you 20k users, it makes the user manager unusable.

The problem is due to a235a2e2296e515ac67e4e618d2c2e4a4fa58e19 : the use of "ajax" param instead of "ajaxsource" is the way to go to use the current API of Datatables, but it completely changes the way parameters are sent to admin/user_list_backend.php. Unfortunately, the user list was apparently working fine, I would have prefered it to completely crash: we would have seen the problem before releasing 2.9.4.

plegall:
I have decided to revert https://github.com/Piwigo/Piwigo/commit/a235a2e2296e515ac67e4e618d2c2e4a4fa58e19 in https://github.com/Piwigo/Piwigo/commit/ab46632a3210b755915c4ecf50d06b6727b1de16

I've considered the facts:
* the original issue, described in #748 has only been reproduced by very few users and is most certainly due to a server configuration (we don't know which one)
* I don't want to switch to GET instead of POST because it will cause huge lines in server log files and I fear the side effects of switching
* we will cleaning use the `ajax` function instead of `ajaxsource` for Piwigo 2.10, see #999

PR #903, opened 2018-07-25
user list: use HTTP GET method for user_list_backend.php

dadap:
Revert to the old Datatables "ajax" API instead of the newer
"ajaxSource" API. This API change was only necessary in order to
supply the pwg_token as POST data, but the user list can be
retrieved without the pwg_token via GET.

### DIFF
--- a/admin/themes/default/template/user_list.tpl
+++ b/admin/themes/default/template/user_list.tpl
@@ -614,14 +614,8 @@ jQuery(document).on('click', '.close-user-details',  function(e) {
     deferRender: true,
     processing: true,
     serverSide: true,
-		serverMethod: "POST",
-    ajax: {
-        url : "admin/user_list_backend.php",
-        type : "POST",
-        data : {
-            pwg_token : pwg_token
-        }
-    },
+		serverMethod: "GET",
+    ajaxSource: "admin/user_list_backend.php",
 		pagingType: "simple",
     language: {
       processing: "{/literal}{'Loading...'|translate|escape:'javascript'}{literal}",


dadap:
I think this should fix #902; pagination and filtering is working for me. Sorry for not noticing that my changes in #866 had broken it.

plegall:
That's the opposite: you're switching back to the old "ajaxsource" instead of the new "ajax" option in Datatables API :-)

If you're not trying to transmit pwg_token, what was the original problem? As I said in #866 

> I don't understand why pwg_token would be necessary to `user_list_backend.php` but it can't hurt :-)

I still don't understand why you wanted to transmit it, but it seemed to fix your issue. Was it actually not the problem? Is your original problem fixed if you switch back to "ajaxsource" and don't transmit pwg_token?

dadap:
Oops. Brain fart when I was writing the commit message. The message was wrong; the change is right. I have force pushed a change with a corrected commit message to my fork.

The original problem was an HTTP 403 when retrieving the user list data from user_list_backend.php. Working from the 403 error code, I guessed there must have been some kind of authentication issue, so I took a look to see how other calls in Piwigo handled authentication. I discovered the pwg_token and as an experiment I tried adding it as POST data to the request. That seemed to fix the problem I was seeing, so I created the PR.

After my first change caused the regression, on your suggestion I tried including the pwg_token as a GET parameter rather than POST data. It worked, but only after changing the method of the request from POST to GET. Once changed to GET, I also tried removing the pwg_token from the GET parameter string, since you seemed to think it was unnecessary, and observed that it was indeed unnecessary, as long as the method is GET.

To summarize:

**Not working (HTTP 403):**

* POST with no pwg_token (This was how it was set up originally, before my first change)

**Working:**

* POST with pwg_token (My first change did this)
* GET with pwg_token
* GET with no pwg_token (this PR)

dadap:
Oops. Brain fart when I was writing the commit message. The message was wrong; the change is right. I have force pushed a change with a corrected commit message to my fork.


The original problem was an HTTP 403 when retrieving the user list data from user_list_backend.php. Working from the 403 error code, I guessed there must have been some kind of authentication issue, so I took a look to see how other calls in Piwigo handled authentication. I discovered the pwg_token and as an experiment I tried adding it as POST data to the request. That seemed to fix the problem I was seeing, so I created the PR.


After my first change caused the regression, on your suggestion I tried including the pwg_token as a GET parameter rather than POST data. It worked, but only after changing the method of the request from POST to GET. Once changed to GET, I also tried removing the pwg_token from the GET parameter string, since you seemed to think it was unnecessary, and observed that it was indeed unnecessary, as long as the method is GET.


To summarize:


**Not working (HTTP 403):**


* POST with no pwg_token (This was how it was set up originally, before my first change)


**Working:**


* POST with pwg_token (My first change did this)
* GET with pwg_token
* GET with no pwg_token (this PR)

plegall:
The drawback of using GET is that you will have huge lines in your web server logs, like this:

> 127.0.0.1 - - [26/Jul/2018:10:45:02 +0200] "GET /pwggit/admin/user_list_backend.php?pwg_token=ef6ee8781a104508d1dc71b20c7e5ab5&sEcho=5&iColumns=7&sColumns=%2C%2C%2C%2C%2C%2C&iDisplayStart=0&iDisplayLength=10&mDataProp_0=0&sSearch_0=&bRegex_0=false&bSearchable_0=true&bSortable_0=true&mDataProp_1=1&sSearch_1=&bRegex_1=false&bSearchable_1=true&bSortable_1=true&mDataProp_2=2&sSearch_2=&bRegex_2=false&bSearchable_2=true&bSortable_2=true&mDataProp_3=3&sSearch_3=&bRegex_3=false&bSearchable_3=true&bSortable_3=true&mDataProp_4=4&sSearch_4=&bRegex_4=false&bSearchable_4=true&bSortable_4=true&mDataProp_5=5&sSearch_5=&bRegex_5=false&bSearchable_5=true&bSortable_5=true&mDataProp_6=6&sSearch_6=&bRegex_6=false&bSearchable_6=true&bSortable_6=true&sSearch=pl&bRegex=false&iSortCol_0=1&sSortDir_0=asc&iSortingCols=1&_=1532593623500 HTTP/1.1" 200 322 "http://localhost/pwggit/admin.php?page=user_list" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; rv:61.0) Gecko/20100101 Firefox/61.0" "-"

I also fear the side effects of changing from POST to GET.

Doesn't it work with POST and:

```
ajaxSource: "admin/user_list_backend.php?pwg_token="+pwg_token,
```

?

plegall:
This pull-request is no longer relevant because the user manager has been completely rewritten and no longer call `user_list_backend.php`

PR #914, opened 2018-08-13
Fix get_absolute_root_url() to detect secure connection

sbias:
if a reverse proxy is used

### DIFF
--- a/include/functions_url.inc.php
+++ b/include/functions_url.inc.php
@@ -53,8 +53,8 @@ function get_absolute_root_url($with_scheme=true)
   {
     $is_https = false;
     if (
-      (isset($_SERVER['HTTPS']) && ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
-      || (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) == 'https')
+      (isset($_SERVER['HTTPS']) and ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
+      or (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) and strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) == 'https')
     ) {
       $is_https = true;
       $url .= 'https://';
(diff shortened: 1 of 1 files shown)


include/functions_url.inc.php
```diff
@@ -52,9 +52,10 @@ function get_absolute_root_url($with_scheme=true)
   if ($with_scheme)
   {
     $is_https = false;
-    if (isset($_SERVER['HTTPS']) &&
-      ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
-    {
+    if (
+      (isset($_SERVER['HTTPS']) && ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
+      || (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) == 'https')
```
samwilson: I don't think this should mix `or` and `||`. I think the Piwigo standard is to favour the former.

choenig:
Hi,

I'd really like to get this applied :-)

I'm new to Piwigo and I almost gave up setting it up. It took me half a day to find this issue as I'm running Piwigo behind a traefik2 https-proxy. Maybe this would also fix #1128. There someone describes the same issues I am having (though he does not talk about a proxy-setup).

Just for completenes: 
My problem is, that i.php returns an 'http'-url when mass-creating thumbnails. My browser rejects that with 'blocked:mixed-content' and so the thumbnail-creation fails.

Br
Christian

jobec:
@samwilson 
Any chance on getting this merged? X-Forwarded-Proto is a defacto standard header for these kind of things.

I personally needed to patch the code while running piwigo inside docker (apache + php) on port 8080 with Caddy in front as reverse proxy for SSL. Piwigo should not try to build the full path to anything and just stick with relative paths and let the browser handle it.

samwilson:
@jobec I'm not a maintainer here, just a random user. My experience has been that PRs are slow to move.

nika1001:
I would also like this issue to be resolved, we use a proxy and its really not a nice solution to patch the code each time a new update comes out.

jobec:
Meanwhile, I bypassed the issue by adding this to my config.inc.php file:

```php
// Support X-Forwarded-Proto header for HTTPS detection
if ( $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' ) {
    $_SERVER['HTTPS'] = 'on';
}
```
That way I don't need to mess around with piwigo's code but still have it respond to the defacto X-Forwarded-Proto header.

GunoH:
> // Support X-Forwarded-Proto header for HTTPS detection
> if ( $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' ) {
>     $_SERVER['HTTPS'] = 'on';
> }

I also had to add `$_SERVER['SERVER_PORT'] = 443;` here, otherwise the `get_absolute_root_url()` function would insert `:80` into the https url.

ambs:
I did not required the SERVER_PORT setting. But the check for X-Forwarded-Proto is a must, and please, should be merged asap.

russell:
the server port is probably also available as the variable `X-Forwarded-Port` so if that is set it could be used to avoid needing to set the port manually to 443

@jobec thank you for that hack, you have saved me some time.

plegall:

Issue #919, opened 2018-08-25
"rank" becomes a mysql reserved keyword in version 8


pay62:
Warning:  [mysql error 1064] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ', status
  FROM piwigo_categories
  WHERE id_uppercat IS NULL
  ORDER BY rank AS' at line 1

SELECT id, name, permalink, dir, **rank**, status
  FROM piwigo_categories
  WHERE id_uppercat IS NULL
  ORDER BY rank ASC
; in D:\www_root\piwigo-2.9.4\include\dblayer\functions_mysqli.inc.php on line 845

**piwigo version: 2.9.4**
**mysql version: 8.0.11**

samwilson:
`rank` is a reserved word as of MySQL 8.0.

I've attempted to find all the places using `rank` and added backticks: #922

There's also the album 'Manual sort order' that gets saved as "rank ASC" and then I think appended to an SQL query when it's loaded. I might be wrong about that though.

UnConundrum:
There are also column named "name" and "status" which are reserved words:

SELECT id, name, permalink, dir, rank, status
 FROM piwigo_categories
 WHERE id_uppercat IS NULL
 ORDER BY rank

samwilson:
@UnConundrum are you getting errors with them? With MySQL 8.0 as well?

I guess best practice would be to quote every column name.

UnConundrum:
Yes, I am.

I guess it’s unrelated, but I’m also getting a JSON error.

UnConundrum:
Any chance this will be worked on in the near future?

samwilson:
I'm not sure.

@plegall might know.

pstimpel:
> There are also column named "name" and "status" which are reserved words:
> 

Add 'section' to it.

plegall:
I gave a general reply about the "reserved keywords in MySQL 8" on https://github.com/Piwigo/Piwigo/issues/996#issuecomment-479804890

Kwaadpepper:
https://github.com/Piwigo/Piwigo/issues/1095

PR #922, opened 2018-08-30
Quote 'rank' field name in SQL

samwilson:
Add backticks to reserved word 'rank' in SQL queries.

Bug: GH #919

### DIFF
--- a/admin/cat_list.php
+++ b/admin/cat_list.php
@@ -349,7 +349,7 @@ $template->assign(array(
 $categories = array();
 
 $query = '
-SELECT id, name, permalink, dir, rank, status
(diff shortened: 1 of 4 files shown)


pay62:
Hello
changing name of the columns would permit to use as usual mysql console to query manually the database.
Olivier

samwilson:
>  changing name of the columns would permit to use as usual mysql console to query manually the database.

That would be a much bigger change, so I think the simpler solution is still worth pursuing (even if it's later updated by renaming the column).

There's no problem with quoting column names in the MySQL console though. I don't think that's an argument for changing the column name.

samwilson:
Does anyone have time to review this patch? :)

PR #925, opened 2018-09-14
Improve permalink generation

plepe:
I found the auto-generated permalinks ugly, so I hacked into the code and made a few improvements:
- allow upper case characters
- replace '.' by '-' (instead of '')
- merge multiple '-' into a single '-'

A second commit chops permalinks at 64 characters. As the database column is limited to 64 characters, this resulted in SQL errors for longer titles.


samwilson:
Will this break old links?

Issue #389, opened 2015-12-30
Multiple formats on API/web upload


plegall:
Now that we have multiple formats on sync (and the appropriate database model), we need to complete it with API/web upload.

PR #929, opened 2018-09-27
Made format subdirectory (i.e. "pwg_format") configurable

joeroe:
Currently the multiple formats feature looks for formats in a hard-coded sub-directory, `pwg_format`, but it's not always desirable to organise files this way. 

This change makes the name of the subdirectory a config variable (`format_dir`), defaulting to `pwg_format` to maintain backwards compatibility. It can be set to `.` to maintain a flat file structure.

### DIFF
--- a/admin/include/functions.php
+++ b/admin/include/functions.php
@@ -610,7 +610,7 @@ function get_fs_directories($path, $recursive = true)
       '.', '..', '.svn',
       'thumbnail', 'pwg_high',
       'pwg_representative',
-      'pwg_format',
+      $conf['format_dir'],
       )
     );
   $exclude_folders = array_flip($exclude_folders);
--- a/admin/site_reader_local.php
+++ b/admin/site_reader_local.php
@@ -112,7 +112,7 @@ function get_elements($path)
       else if (is_dir($path.'/'.$node)
                and $node != 'pwg_high'
                and $node != 'pwg_representative'
-               and $node != 'pwg_format'
+               and $node != $conf['format_dir']
                and $node != 'thumbnail' )
       {
         $subdirs[] = $node;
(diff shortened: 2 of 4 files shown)


joeroe:
Another small change for maintaining tidy file systems: ignoring unix-style hidden files, e.g. the many empty ._ files created by MacOS.

plegall:
Hi @joeroe 

Your code seems fine, but I would like to better understand why it's fine to have hard-coded `pwg_representative` and not `pwg_format`

joeroe:
Thank you for the quick response @plegall. The simple answer is that I don't use `pwg_representative`, only `pwg_format`, so I only looked at that. I think it would be nice for both to be configurable.

To explain a little: I'm using Piwigo as a friendly interface for a preexisting gallery of scientific images that are linked to other databases. Therefore I'm trying to avoid changing the directory structure or inserting application-specific files, as much as possible.

joeroe:
@plegall Is it possible to get an update on whether this will be merged into master? If not I would like to investigate making it a plugin, as each update is currently breaking my file structure.

Arnie97:
This looks promising. I'd like my raw images to be kept in the same directory as JPEGs - which is also the choice of many camera manufacturers. Any updates?

joeroe:
@plegall Sorry to be a bother, but any updates on this? The latest update has once again broken my formats. Would it help if I also made `pwg_representative ` configurable and added it to the pull request?

plegall:
Hi @joeroe 

> @plegall Would it help if I also made `pwg_representative ` configurable and added it to the pull request?

Yes please, do the same for `pwg_representative` and I'll merge the pull-request.

Warning: with #389 we have added some occurences of `pwg_format` in core PHP code.

joeroe:
@plegall Is it possible to merge this without the change to `pwg_representative`? I don't need the latter and at this point it's been so long since I touched the Piwigo codebase I wouldn't know how to go about adding it.

PR #936, opened 2018-10-24
Switch if-statement for readability.

lennartbecker:


plegall:
Hi @lennartbecker 

The `if ('value' == $var)` is perfectly written "on purpose". Why is that? because if you write `if ($var == 'value')` you may write `if ($var = 'value')` by mistake and this will always be true... Instead if you write `if ('value' = $var)` by mistake you will immediately get an error.

samwilson:
https://en.wikipedia.org/wiki/Yoda_conditions

samwilson:
I imagine the intention is that all applicable conditionals in Piwigo will one day conform to that style. But software is always a _process_... :-)

plegall:
>  But why is it only written like that sometimes?

As far as you read code I wrote, it should always be in "yoda" style :-) but I'm far from being the only one who code Piwigo!

samwilson:
Maybe this PR should be changed to update the non-yoda style conditionals?

Issue #945, opened 2018-11-25
pwg_log is not checking if given $page['section'] is part of the enum in the table


pstimpel:
There are plugins out there setting $page['section'] to something own, like the the Guestbook plugin or the Contactform.

This causes problems with pwg_log. pwg_log only checks of a section was given, and if so it pushed the new row to the database. Now mysql.history.section is an enum with the members 

'categories','tags','search','list','favorites','most_visited','best_rated','recent_pics','recent_cats','additional_page'

This means, the insert fails and prints an ugly warning to the Piwigo page. 

Of course IMO it is bad style to abuse section this way, maybe this is based on historic grounds. However, the issue is there and is live. I guess people will not recognize it since it is just a warning which is not shown in most productive environments.

`Warning:  [mysql error 1265] Data truncated for column 'section' at row 1

INSERT INTO piwigo_history
  (
    date,
    time,
    user_id,
    IP,
    section,
    category_id,
    image_id,
    image_type,
    format_id,
    auth_key_id,
    tag_ids
  )
  VALUES
  (
    CURRENT_DATE,
    CURRENT_TIME,
    2,
    '192.168.0.22',
    'guestbook',
    NULL,
    NULL,
    NULL,
    NULL,
    NULL,
    NULL
  )
; in /media/sf_piwigo/include/dblayer/functions_mysqli.inc.php on line 845`

But it is still a bug. In my opinion this should be fixed by a check inside pwg_log, if the given section is part of the enum. If it is not, just null it and pwg_log will go on. By doing so, all these plugins don't have to be rewritten.

BR,
Peter

pstimpel:
Wrote a fix and made a pull request, maybe this helps.

PR #946, opened 2018-11-27
Wrong handling of strange section fields in pwg_log

pstimpel:
If plugin developers add their own sections, Piwigo tries to put them into history table. Since the section column is an enum, php issued a warning. 

Addresses https://github.com/Piwigo/Piwigo/issues/945


plegall:
I'm merging, but I'm going to improve a little :-)

PR #947, opened 2018-11-27
Addresses conflicts when using Mysql8 in pwg_log

pstimpel:
In Mysql8 reserved words in sql statements have to go in backticks. Of course this fix is addressing only one of many occurences accross the core, but maybe it can be a start of making Piwigo Mysql8 comptatible.


Kwaadpepper:
please see https://github.com/Piwigo/Piwigo/issues/1095

PR #952, opened 2018-12-26
Update JShrink for php73 support

tkuther:
Note: requires namespaces

### DIFF
--- a/include/jshrink.class.php
+++ b/include/jshrink.class.php
@@ -1,469 +1,592 @@
 <?php
+/*
+ * This file is part of the JShrink package.
+ *
+ * (c) Robert Hafner <tedivm@tedivm.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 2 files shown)


samwilson:
I wonder if there's any chance of Piwigo using Composer, and loading JShrink from [tedivm/jshrink](https://packagist.org/packages/tedivm/jshrink)?

(I did try adding composer.json in #895, only for dev requirements, but there's been no movement on that front.)

Teatek:
There is a problem for this issue because the required PHP version for JShrink 1.3.1 is 5.6 and piwigo 2.9.4 only require 5.3 or newer.
Would it be possible to isolate the specific change for this issue ?

tkuther:
Why keep support for PHP versions that EOL'ed years ago? Even 5.6 is EOL meanwhile.

plegall:
PR has already been updated. I close this PR

PR #953, opened 2019-01-08
adding trigger tag delete (issue 854)

Teatek:


plegall:
Can you also update `tools/triggers_list.php`

PR #954, opened 2019-01-08
trigger group delete (issue 855)

Teatek:


plegall:
Can you also update `tools/triggers_list.php` ?

PR #955, opened 2019-01-09
Issue 917 album photo sort order

Teatek:


plegall:
@Teatek please apply the same algorithm to image_order for `pwg.categories.getAdminList` as you do for `pwg.categories.getList`

Issue #278, opened 2015-07-25
[security] prevent discovering hidden tags from URL


plegall:
**Reported by plg on 5 Oct 2014 18:57**

Even if tags are not shown on tags.php, it is easy to increment the tag id in URL to disover all tags (even if the page shows no photo)

[Piwigo Bugtracker #3155](http://piwigo.org/bugs/view.php?id=3155)

plegall:
commit 29f938acba2b11d3b4a29abbe52c032495a005b9 by @Teatek

PR #962, opened 2019-01-16
display login page if there is no images at the tag (URL)

Teatek:


plegall:
This pull-request is related to #278

Issue #698, opened 2017-06-01
show_metadata=true vs. noindex,nofollow logic in picture.php


tkuther:
In local/config/config.inc.php a site admin can enable displaying metadata by default. Also theme devolpers might decide to set this to true via pwg_set_session_var.

If set to true, the logic in picture.php adds the robots noindex,nofollow meta tag, which basically disables indexing of the picture pages. This does make sense if show_metadata=false by default, so the picture page doesn't get indexed twice following the "toggle metadata" button.

If set to true by default, that logic should acually be reversed. Index by default (with metadata then), noindex,nofollow when following the link to the page without metadata. Implementing this sounds overly complicated, though. We have a semi-hidden config.inc.php option, user interaction with a button press , maybe a forced session var from a theme, etc.

To keep things simple, a site admin should be able to set this in the admin panel (show metadata always, never, only for authenticated users) and the test in picture.php should be removed. Theme developers still can implement the toggle button using jQuery show/hide, CSS or similar.

References:
http://piwigo.org/forum/viewtopic.php?id=28008
https://github.com/tkuther/piwigo-bootstrap-darkroom/issues/64

reisi007:
Hi,
Any update on this? My picture pages are not indexed....

reisi007:
@tkuther @flop25 @plegall : Can you look at my pull request and give me input on the change. Tested (and keeping it) using bootstrap darkroom and the new config set to true -> Google sees my picture pages now ;)

How shall we continue?

hesyifei:
Any update on this?

Photonoxx:
Asking the same thing : Any update on this ?

I found some message about this matter and Stripped theme dating from 2011.
May be it could be time to test the commit ?
Thanks

timfragments:
Any update on this topic ? Thanks

HWFord:
Following this issue we realised that because some themes always display metadata so the no index no follow was always added. Rather than base the adding of the meta tag on the show_metadata variable stored in the session, we only look at the url, so the no follow is not always present for the themes that decide to always show metadata

PR #970, opened 2019-01-24
Search engine won't index picture page | Correct logic and makes issue #698 fixable by themes / user

reisi007:
To discuss:

Should this be set by themes automatically or should there be an UI toggle (I might not be able to do this) to make this easily changeable by user.

Although this seems like a lot of work, I am for themes fixing this issue. If theme does not fix it, no change in behaviour is expected. So this seems like an good idea given this issue has been over a year in prio highest.


hesyifei:
Any update on this? (Also just linking it to #698)

PR #971, opened 2019-01-25
Issue 863 error delete

Teatek:


plegall:
The description of method `pwg.categories.deleteRepresentative` also needs to be updated:

> Deletes the album thumbnail. Only possible if $conf['allow_random_representative'] or if the album has no direct photos.

Issue #996, opened 2019-03-31
administration page failed loading when piwigo is installad without any table prefix


pay62:
**Versions**
Release : head of master branch
MySQL: 8.0.11
PHP: 7.2.4

**Step to reproduce**
1 Install piwigo on a newly created database without any prefix for table name
2 Add a picture
3 With menu on left got to administration page

**Results**
:heavy_check_mark:Step 1: OK
:heavy_check_mark:Step 2: OK
 :x:Step 3: KO. Following error message appeared in the browser

Piwigo a rencontré une erreur non récupérable
[mysql error 1064] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'groups' at line 2

SELECT COUNT(*)
  FROM groups
;

#1	my_error D:\www_root\piwigo_src\include\dblayer\functions_mysqli.inc.php(147)
#2	pwg_query D:\www_root\piwigo_src\admin\intro.php(190)
#3	include D:\www_root\piwigo_src\admin.php(324)

plegall:
OK, so "groups" is now a reserved keyword in MySQL 8 :-/ gonna have to escape it with backtits!

samwilson:
PR #922 is related (for the `rank` keyword).

plegall:
@samwilson @pstimpel my opinion right now, is that Piwigo needs to escape column names with backtits everywhere, which has less "risky" impact than renaming all problematic columns.

Considering MySQL 8 is not yet very common, I think we can plan this important change for Piwigo 2.10 and not for 2.9. Do you both agree with that?

samwilson:
Yep most definitely. :-)

I don't even really think there's much wrong with having column names the same as keywords. Better to not to, certainly, but no need to be a high priority to fix.

(And by the way, it's `backticks`, with a `ck`...)

pstimpel:
I agree with the backticks, and with your timeline as well. Keep in mind we do not know about Oracles plans for Mysql. Maybe you will create kind of a sleeping nightmare. But yes, for th moment this should be OK

plegall:
> (And by the way, it's `backticks`, with a `ck`...)

LOL, I don't want to know what "backtit" might mean ;-) thank you for telling me it's backtick :-D

pay62:
> @samwilson @pstimpel my opinion right now, is that Piwigo needs to escape column names with backtits everywhere, which has less "risky" impact than renaming all problematic columns.
> 
[...]

PR #997, opened 2019-03-31
issue #919 replace rank column in database

pay62:
rank is replaced by ima_rank for image_category table and by cat_rank
for categories table
Fix was tested successfully on tag 2.9.4 and also on head of master branch
Installation of piwigo is OK
Photos notation is OK
Display photos with best notation is OK

### DIFF
--- a/admin/cat_list.php
+++ b/admin/cat_list.php
@@ -309,7 +309,7 @@ $template->assign(array(
 $categories = array();
 
 $query = '
-SELECT id, name, permalink, dir, rank, status
+SELECT id, name, permalink, dir, cat_rank, status
   FROM '.CATEGORIES_TABLE;
 if (!isset($_GET['parent_id']))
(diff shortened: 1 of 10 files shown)


samwilson:
Does this actually alter the column names anywhere? It looks like it just replaces them, and modifies the initial DB installation, but doesn't do any migration.

What's wrong with just quoting the `rank` keyword (e.g. #922)?

pay62:
Migration is not implemented.
I have to push one more commit (remaining one rank somewhere)
From my point of view quoting could be fine but make database work painful since you have to quote column name while making your sql queries.
I think also it's a better code to avoid reserved keywords for column name.

plegall:
As said in https://github.com/Piwigo/Piwigo/issues/996#issuecomment-479804890 for now we simply escape problematic column name with backticks

PR #1030, opened 2019-06-07
specify which table to order by

gnowxilef:
this is probably because of mysql 8.


gnowxilef:
ignore me. this is the mysql 8 "rank" issue.

Issue #1061, opened 2019-08-14
Fails to load theme if user theme has been uninstalled


samwilson:
If a user has a theme set which is subsequently removed from the filesystem (i.e. still is the active theme in the user's preferences), the default theme fails to load and the following warning is issued:

```
( ! )  Warning: include(/themeconf.inc.php): failed to open stream: No such  file or directory in .../piwigo/include/template.class.php on  line 1141
--
1 | 0.0001 | 402880 | {main}(  ) | .../index.php:0
2 | 0.0002 | 403696 | include_once( '.../piwigo/include/common.inc.php' ) | .../index.php:11
3 | 0.0112 | 773840 | Template->__construct(  ) | .../common.inc.php:218
4 | 0.0119 | 793256 | Template->set_theme(  ) | .../template.class.php:128
5 | 0.0119 | 793824 | Template->load_themeconf(  ) | .../template.class.php:169
```

This is with current master (63bbba8e36e).

plegall:
Nicely done, thank you @samwilson

samwilson:
No worries, thanks for the quick merge! :-)

PR #1062, opened 2019-08-14
Use default theme if loaded one is not found

samwilson:
When loading a user's theme, make sure it's installed. Also, when
determining the default theme, account for the situation in which
zero themes are installed (and use 'default' then).

Refs #1061


plegall:
Very good, tested and approved. It was a little pain every time I had this problem. Better have a clean fallback than an ugly error.

PR #1071, opened 2019-09-06
Tags : use math inter char ∩ instead of +

7tonin:
What about changing this char ? This helps not to confuse mathematical set theory with mathematical symbol operator. Further more, from set theory, union (+) and intersection (.) can be translated to plus (+) and times (.) operations.

### DIFF
--- a/include/functions_html.inc.php
+++ b/include/functions_html.inc.php
@@ -403,7 +403,7 @@ function get_tags_content_title()
 
   for ($i=0; $i<count($page['tags']); $i++)
   {
-    $title.= $i>0 ? ' + ' : '';
+    $title.= $i>0 ? ' ∩&#x202F;' : '';
 
     $title.=
       '<a href="'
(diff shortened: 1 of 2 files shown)


7tonin:
One may prefer & (and) instead of +

7tonin:
[b]A1+A2[/b] ? https://piwigo.org/forum/viewtopic.php?pid=179351#p179351
After testing, I understand you actually mean pictures belonging to [b]A1 and A2[/b], or in Boolean algebraic operations  [b]A1 & A2[/b], or in set theory symbolism  [b]A1 ∩ A2[/b], or in maths symbolism  [b]A1 × A2[/b], and [b]A1 . A2[/b] as well.
cf. issue "Tags : use math inter char ∩ instead of +" https://github.com/Piwigo/Piwigo/pull/1071
or, |, ∪,  + 
and, &, ∩, ×, ., ⋅, *

Noticed that "SmartAlbums" extension would be able to deal with ∪ union https://fr.piwigo.org/ext/extension_view.php?eid=544
Can filter with these options : (all | one of | none of | only) these (albums | tags)

PR #1082, opened 2019-09-30
(Fix #359) mysqli docs use empty string for no db name

### LINKED ISSUE #359: database connection error with PHP 5.6

PatrickCronin:
[PHP docs for mysqli's constructor](https://www.php.net/manual/en/mysqli.construct.php) show that the default value for `$dbname` is an empty string. So, if we want to use parameters after `$dbname` in the call without specifying a database name, we should be using an empty string for `$dbname` instead of a `NULL`.

### DIFF
--- a/include/dblayer/functions_mysqli.inc.php
+++ b/include/dblayer/functions_mysqli.inc.php
@@ -47,7 +47,7 @@ function pwg_db_connect($host, $user, $password, $database)
     list($host, $port) = explode(':', $host);
   }
 
-  $dbname = null;
+  $dbname = '';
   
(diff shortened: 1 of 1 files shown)


plegall:
It has never seemed to be a problem but why not (if it breaks nothing)

PR #1103, opened 2019-10-16
Fix jshrink file corruption under PHP 7.3

rlerdorf:
In PHP 7.3 you will get a warning from strpos() when you pass it
a non-string needle. Since getChar() will return false in some
circumstances, this->a and this->b will be non-strings sometimes
and the warning can end up in the resulting js output causing
syntax errors. In my case it meant that the easyRotate plugin
didn't work because of this bug. With this fix it works fine.

### DIFF
--- a/include/jshrink.class.php
+++ b/include/jshrink.class.php
@@ -192,7 +192,7 @@ class JShrink_Minifier
 				case "\n":
 					// if the next line is something that can't stand alone
                     // preserve the newline
-					if($this->b !== false && strpos('(-+{[@', $this->b) !== false)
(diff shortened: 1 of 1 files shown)


rlerdorf:
I actually meant for that 2nd patch in this PR to be a separate pull request, but I don't have write access to fix it. I suggest considering them separately.

plegall:
I would like to close this pull-request because jshrink has been updated several times and is currently compatible with, at least, PHP 8.0.

Concerning the Phan, can you make a clean separate PR for this? I admit it's scary huge. I also wonder how useful it is. I mainly see some type cast. What do you think?

PR #1104, opened 2019-10-16
Fix Finnish plurals in Admin dashboard stats

iiska:
Use partitive case after number values:
https://en.wikipedia.org/wiki/Partitive_case


plegall:
Hi @iiska 

The changes in language files must be made through piwigo.org/translate : do you have an account there?

PR #1117, opened 2019-11-06
Quote mysql keyword groups

### LINKED ISSUE #1095: mysql 8 groups is a reserved keyword
When accessing User config page in admin i get a sql error

I tried to create a user, then looked at XHR request, it dumps a SQL statement, this is a issue also because it reveals SQL structure. Debug options should be hidden in production.

    SELECT
    user_id,
    GROUP_CONCAT(name ORDER BY name SEPARATOR ", ") AS groups
      FROM piwigo_user_group
    JOIN `piwigo_groups` ON id = group_id
      WHERE user_id IN (1,2,3)
      GROUP BY user_id

```AS groups```   => groups is a reserved keyword in mysql 8 query fails

https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-G

    GROUPS (R); added in 8.0.2 (reserved)

image

EDIT: updated the SQL query forgot to correct as it is in piwigo files  (group => groups)

thstieler:
Quote groups, which is a reserved keyword in MySQL 8
Fixes  #1095

### DIFF
--- a/admin/user_list_backend.php
+++ b/admin/user_list_backend.php
@@ -201,7 +201,7 @@ if (count($user_ids) > 0)
   $query = '
 SELECT
     user_id,
-    GROUP_CONCAT(name ORDER BY name SEPARATOR ", ") AS groups
+    GROUP_CONCAT(name ORDER BY name SEPARATOR ", ") AS `groups`
   FROM '.USER_GROUP_TABLE.'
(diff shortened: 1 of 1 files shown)


gnowxilef:
can this be merged please?

sridharb1:
Can this be merged to 2.10.1?

PR #1131, opened 2019-12-02
fix: missing l10n item

Arnie97:


admin/picture_modify.php
```diff
@@ -284,7 +284,7 @@
   'file' => l10n('Original file : %s', $row['file']),
   'add_date' => l10n('Posted %s on %s', time_since($row['date_available'], 'year'), format_date($row['date_available'], array('day', 'month', 'year'))),
   'added_by' => l10n('Added by %s', $row['added_by']),
-  'size' => $row['width'].'&times;'.$row['height'].' pixels, '.sprintf('%.2f', $row['filesize']/1024).'MB',
+  'size' => l10n('%s pixels, %.2f MB', $row['width'].'&times;'.$row['height'], $row['filesize']/1024),
```
samwilson, line 287: The `%s pixels, %.2f MB` message should also be added to `language/en_UK/admin.lang.php`, I think.

plegall:
This is a good code suggestion. I have to merge it (and add the new language string in language files, at least in `en_UK`)

Issue #1028, opened 2019-06-03
zh lang problem,an EIGHT yesrs old bug.


edouardlicn:
https://cn.piwigo.org/forum/viewtopic.php?pid=112239#p112239

See above,about 8 years ago I use piwigo and found a bug,while I install zh_tw or zh_hk,if I set default language zh_cn it wont work,the language will be zh_tw or zh_hk.AFTER 8 YEARS I
'm back and the bug still be there.Any one can check it?

Priviosly I found it store the setting in db so after the zh_tw or zh_hk installed it will change the default lang setting.I don't know if it still be there.

EddyLB:
@Piwigo-TranslationTeam , @plegall , would it be possible to finally fix that bug?

plegall:
fixed thanks to #1132

PR #1132, opened 2019-12-03
fixes #680, #875 and #1028: improve detection for browser preferred languages

### LINKED ISSUE #680: Detect the language of the visitor thanks to the browser
Detect the user's language and then redirect it to the language of the user.
Accept-Language : https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/Accept-Language

Arnie97:
The `get_browser_language()` feature was introduced near a decade ago in https://github.com/piwigo/Piwigo/commit/6d8c172ee2e06b2f9e14f9189273403a990cdefa -- in which days there were only 3 translations other than `en_UK`, and hence a more complicated parser for the `Accept-Language` header would simply be an overkill.

Time flies and Piwigo has become available in 70 different languages around the world. The simple `substr($HEADER, 0, 2)` trick is no longer capable to cover all the use cases, and I think it's time to move on to a real parser. I'm a newbie in PHP, so feel free to post any ideas or suggestions you might have here.

Some test cases:

| browser preference | current piwigo behavior | proposed patch |
| --- | --- | --- |
| en-US,en;q=0.9 | en_GB | en_US |
| be-BY,ru-RU;q=0.9 | en_UK * | ru_RU |
| zh-CN | zh_TW | zh_CN |
| ko-KR | kok_IN | ko_KR |
| ko-KP | kok_IN | ko_KR |

\* `$PHPWG_DEFAULT_LANGUAGE`

### DIFF
--- a/include/functions_user.inc.php
+++ b/include/functions_user.inc.php
@@ -793,20 +793,76 @@ function get_default_language()
 
 /**
  * Tries to find the browser language among available languages.
- * @todo : try to match 'fr_CA' before 'fr'
  *
  * @return string
  */
 function get_browser_language()
 {
-  $browser_language = substr(@$_SERVER["HTTP_ACCEPT_LANGUAGE"], 0, 2);
+  $language_header = @$_SERVER['HTTP_ACCEPT_LANGUAGE'];
+  if ($language_header == '')
+  {
+    return false;
+  }
+
+  // case insensitive match
+  // 'en-US;q=0.9, fr-CH, kok-IN;q=0.7' => 'en_us;q=0.9, fr_ch, kok_in;q=0.7'
+  $language_header = strtolower(str_replace("-", "_", $language_header));
+  $match_pattern = '/(([a-z]{1,8})(?:_[a-z0-9]{1,8})*)\s*(?:;\s*q\s*=\s*([01](?:\.[0-9]{0,3})?))?/';
+  $matches = null;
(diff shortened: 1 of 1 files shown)


plegall:
Tested and it does indeed a much better job, compared to the old `get_browser_language()` function. I was worried it would have bad performances, but I checked that too:

```
echo 'HTTP_ACCEPT_LANGUAGE : '.$_SERVER['HTTP_ACCEPT_LANGUAGE'].'<br>';

$start_time = get_moment();
for ($i=0; $i < 1000; $i++)
{
  get_browser_language();
}
echo 'get_browser_language : '.get_browser_language().' ('.get_elapsed_time($start_time, get_moment()).')<br>';

$start_time = get_moment();
for ($i=0; $i < 1000; $i++)
{
  get_browser_language2();
}
echo 'get_browser_language2 : '.get_browser_language2().' ('.get_elapsed_time($start_time, get_moment()).')<br>';
```

And the output is:

```
HTTP_ACCEPT_LANGUAGE : en-US,en;q=0.9,ko;q=0.8,zh;q=0.7,en-GB;q=0.6,de;q=0.5,fr-FR;q=0.4,fr;q=0.3
get_browser_language : en_GB (0.193 s)
get_browser_language2 : en_US (0.232 s)
```

For a 1k iterations loop, the difference is acceptable :-)

Issue #1160, opened 2020-02-15
Add ability to expand the display categories of tags in a theme


oOjayOo:
I would like to add a category to the keyword overview in my theme. Unfortunately, this is not possible, because all deviations are intercepted in php.

plegall:
Hi @oOjayOo 

Can you give more details/examples? I don't see what's the requested feature here :-/ Isn't it what plugin Tag Groups does? See https://piwigo.org/ext/extension_view.php?eid=781

PR #1161, opened 2020-02-15
issue #1160 Add ability to expand the display categories of tags in a…

oOjayOo:
… theme


plegall:
There are so many "code style" changes, that I don't even see the real change. Can you revert the "code style" changes (in a new commit of the same pull-request) and focus on real changes? (in order to merge/close pull-requests, I will close it by the end of august 2022 if no update given)

PR #1164, opened 2020-02-20
add Perl module Mojo::Piwigo::Client and demos

mbethke:
I'm glad to see Piwigo has a lot of Perl tooling, but TBH the whole collection has a bit of a 90s feel to it; we've had much simpler and more flexible tools than the venerable workhorse `LWP` for a couple of years now, such as [Mojo](https://metacpan.org/pod/Mojo::UserAgent). So when I started hacking on `piwigo_upload.pl` to fix some weird server errors I'd been getting, I quickly decided to write a new module based on that, `Mojo::Piwigo::Client`.
Here's a first draft that does a few things plus two demos roughly equivalent to some included tools.
Only a few methods are implemented, but I'm not sure it's worth adding them all as you can always use `call_res` to call any method and receive the result as a Perl data structure. Ideas welcome--autogenerating methods from the reflection API would work, too ;)

Additionally, there's a trivial little config module, `Mojo::Piwigo::Client::Config`, that allows you to stick your credentials in a YAML file and forget about them instead of specifying them on the commandline every time, while still making it possible to override the store ones from the commandline when needed.

### DIFF
--- a/tools/lib/Mojo/Piwigo/Client.pm
+++ b/tools/lib/Mojo/Piwigo/Client.pm
@@ -1,0 +1,332 @@
+# ABSTRACT: Client for Piwigo's Web Service API
+use 5.024; 
+
+package Mojo::Piwigo::Client {
+    use version 0.77; our $VERSION = version->declare("v0.0.1");
+    use Mojo::Base -base;
+    use Mojo::UserAgent;
+    use Mojo::Exception;
+    use Path::Tiny;
+    use POSIX 'ceil';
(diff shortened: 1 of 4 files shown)


plegall:
Hi @mbethke 

I think your scripts are interesting but I also don't think they belong to Piwigo core bundle. I suggest (because I don't want this piece of code to be lost, really) that you package your own repository for those scripts and propose them as a tool on https://piwigo.org/ext/

What do you think?

Issue #408, opened 2016-01-27
PHP7, Catchable fatal error: session_regenerate_id(): Failed to create(read) session ID


plegall:
http://piwigo.org/forum/viewtopic.php?id=26239

This is a PHP 7 bug, that should be fixed in version 7.0.3 (release date = 2016-02-07), see https://github.com/php/php-src/blob/PHP-7.0/NEWS#L70

fcoiffie:
With PHP 7.0.3, the problem still occurs... (PHP 7.0.3-1~dotdeb+8.1 on Debian Jessie)

plegall:
@fcoiffie thank you for this report

@nikrou also reported a problem with PHP 7.0.2 and proposed a fixed. Can you create a PR?

nikrou:
Of course I can but I cannot already test with PHP 7.0.3

plegall:
Unfortunately, even with fix from @nikrou, I still have error on PHP 7.0.3

Catchable fatal error: session_regenerate_id(): Failed to create(read) session ID: user (path: /var/lib/php/sessions) in .../include/functions_user.inc.php on line 953

I'm going to remove the 7.0.3 test.

plegall:
@fcoiffie can you make a new try with https://github.com/Piwigo/Piwigo/commit/e71b61aee1b3e077899bb8276398f4c705586e51 ?

fcoiffie:
I manually applied this patch on my Piwigo version and it works : no error. (But I can't say if it's safe...)

plegall:
When session_regenerate_id() will be fixed on PHP 7, we will use it again.

PR #1178, opened 2020-04-22
Fix version comparison for session ID regen

takuy:
The comparison should be the other way around for PHP version comparison. session_regenerate_id() was available in 7.0.0 and newer versions. version_compare returns 0 on equality by default, and 1 when the first argument is greater than the second argument.

### DIFF
--- a/include/functions_user.inc.php
+++ b/include/functions_user.inc.php
@@ -936,10 +936,7 @@ function log_user($user_id, $remember_me)
   if ( session_id()!="" )
   { // we regenerate the session for security reasons
     // see http://www.acros.si/papers/session_fixation.pdf
-    if (version_compare(PHP_VERSION, '7') >= 0)
-    {
-      session_regenerate_id(true);
-    }
+    session_regenerate_id(true);
   }
   else
   {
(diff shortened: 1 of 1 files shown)


plegall:
I think it's a bit more complicated, did you check:

- f04e4d8b4fe338680f5931bc1609b9201c6cd023
- 0909717c37cdf2ddadcddaf92d610f5243ec3c17
- 08fe998dd10fbf3fe79e8722df891b9496228b4a

takuy:
Yes, I looked at those. Not sure what causes the errors that were reported (memcached? redis?). But officially session_regenerate_id() was available in 7.0.0. If you look here - https://3v4l.org/Y85gv - you can see session_regenerate_id() works without error. 

Currently, Piwigo only regenerates the Session ID if the version is OLDER than 7.0.0, which is not secure at all as PHP 5.6 is officially deprecated.  Session IDs are not regenerated on login for PHP 7 or greater.

plegall:
> officially session_regenerate_id() was available in 7.0.0.

And obviously #408 it was not working on Piwigo :-/

> If you look here - https://3v4l.org/Y85gv - you can see session_regenerate_id() works without error.

And it says it works for PHP 7.0.3 while many reports from users told us it was not working. Maybe the way we handle sessions in Piwigo (in database, instead of files) was causing the issue.

But I agree, on master we can completely remove the test and do the `session_regenerate_id` whatever the PHP version is. We'll see if there is feedback from beta testers.

Do you agree with that? Can you remove the condition?

PR #1191, opened 2020-06-07
Sort filters by localized name

samwilson:
The `UC_name_compare` function isn't defined, but was being used
to sort the prefilter list. This change adds a anonymous function
to do the sort comparison (on the filters' NAME attribute).


gasp (approved):

samwilson:
Thanks!

PR #1268, opened 2020-11-21
Fix stored xss 

abhaysingh1998:
Saw a Stored XSS while creating an album with the name as that of the following payload at the following endpoint,

Endpoint:- */admin.php?page=cat_list*
Payload:- *<script>alert('xss')</script>*

POC Demo:- [Demo](https://drive.google.com/file/d/1B3yip5wqYOXUESLCANWnuY45nBMKuysK/view?usp=sharing)

Fixed it by wrapping *virtual_name* post parameter inside strip_tags function. 
Please review.

### DIFF
--- a/admin/cat_list.php
+++ b/admin/cat_list.php
@@ -159,7 +159,7 @@ if (isset($_GET['delete']) and is_numeric($_GET['delete']))
 elseif (isset($_POST['submitAdd']))
 {
   $output_create = create_virtual_category(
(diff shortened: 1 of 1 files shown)


plegall:
It's not a security issue. Only admins can create/edit albums and admins can do much more than inserting javascript in an album name.

Furthermore, we already have a setting `$conf['allow_html_descriptions']` : why not using it to `strip_tags` on author and name?

Issue #323, opened 2015-07-25
[synchronization] Normalize UTF-8 filenames before synchronization


plegall:
**Reported by mindhaq on 9 Apr 2015 14:51**

**Version:** 2.7.4

Filenames scanned during synchronization should be UTF-8 normalized to the precomposed format using Normalizer. At least when default encoding is utf-8 and PHP 5.3 is available.

http://us2.php.net/manual/en/class.normalizer.php

**Additional information:** UTF-8 allows two different ways to encode umlaut special characters like the german ä, ü, ä. For technical background on that see e.g. the following:

http://stackoverflow.com/questions/1089966/utf8-filenames-in-php-and-different-unicode-encodings

I upload images from my Mac to my Linux server via SSH, and some of them contain Umlaut characters. For those files to be imported, I already changed the regex to the following:

$conf[&#039;sync_chars_regex&#039;] = &#039;/^[\s,.\&#039;\pL0-9-_.]+$/u&#039;;

Which works fine for files with Umlaut I create directly on the server.

On the Mac, the Umlauts ä are encoded in the decomposed format 0x61cc88. When I upload them, they keep that filename. This is no problem for Linux, the names are displayed just fine. However, PHPs regex parser does not match those decomposed characters with the \pL character group (which it should). So it would be great if those filenames would be normalized before matching.

[Piwigo Bugtracker #3218](http://piwigo.org/bugs/view.php?id=3218)

fri-K:
Is it possible to move $conf['sync_chars_regex'] = '/^[a-zA-Z0-9-_.]+$/'; line to configuration file from admin/site_update.php:763? Or at least use some variable from config file?
After each piwigo update I must add my own regex over and over again because this files is always overwritten.

flop25:
Did you read http://piwigo.org/doc/doku.php?id=user_documentation:use:features:conf_locale ?

fri-K:
@flop25 It looks like it should work, I'll see on next update. Thanks for the tip :)

Issue #218, opened 2015-07-25
[image processing] an ampersand in album names prevents thumbnail generation


plegall:
**Reported by tkl on 28 Feb 2013 23:20**

**Version:** 2.4.7

In albums that have names containing an ampersand, any image created from the original is never shown. Creating directories with ampersands in the name is fine, given that in local/config.inc.php the ampersand, like so: \&amp; is listed in $conf[&#039;sync_chars_regex&#039;]. The thumbnails in these albums are never shown, because in the thumbnail directory, _data/i/galleries/, directories with ampersand are not created.

**Steps to reproduce:** mkdir galleries/&#039;juliet &amp; romeo&#039;
cp forensicevidence/detailabdomen.jpg galleries/&#039;juliet &amp; romeo&#039;

In piwigo, go to adminstration -&gt; tools -&gt; synchronize
add directories and files

Go back to the gallery: Any size not the original does not show up; In the _data/i/galleries directory juliet &amp; romeo is missing.

[Piwigo Bugtracker #2853](http://piwigo.org/bugs/view.php?id=2853)

Issue #201, opened 2015-07-25
[synchronization] Unable to synchronize directories starting with special names


plegall:
**Reported by gospogied on 2 Oct 2012 07:29**

**Version:** 2.4.4

basename() is locale dependent so before it we must setlocale for special strings.

**Steps to reproduce:** 1. Edit 
change in file include/config_default.inc.php line 743
 $conf[&#039;sync_chars_regex&#039;] = &#039;/^[a-zA-Z0-9-_.]+$/&#039;;
to
 $conf[&#039;sync_chars_regex&#039;] = &#039;/^[a-zA-Z??ó????????Ó??????0-9-_. (),%+-~#@!*]+$/&#039;;
2. Create dir:
mkdir /PATH/TO/PIWIGO/galleries/?uk /PATH/TO/PIWIGO/galleries/luk /PATH/TO/PIWIGO/galleries/ku?
3. Synchronize. 
Now You can see following albums in piwigo: uk luk ku?
So it ignores special characters in first positions.

The problem is in admin/site_update.php line 210.
To solve it you must add one line before it, for my language it is:

setlocale(LC_ALL, &#039;pl_PL.UTF-8&#039;);
$dir = basename($fulldir);

**Additional information:** Fixed by hand, but should be automated and probably basenames in other places also should be setlocaled first.

[Piwigo Bugtracker #2758](http://piwigo.org/bugs/view.php?id=2758)

plegall:
**Comment posted by gospogied on 2 Oct 2012 08:15**

It seems that bugtracker doesn&#039;t like locales, according to ????

plegall:
**Comment posted by m_alien on 31 Oct 2014 07:41**

in piwigo 2.7.1 same broblem, and the same solution:

The problem is in admin/site_update.php line 210.
 To solve it you must add one line before it, for my language it is:

 setlocale(LC_ALL, &#039;ru_RU.UTF-8&#039;);
 $dir = basename($fulldir);


and in i.php I add  

setlocale(LC_ALL, &#039;ru_RU.UTF-8&#039;);

PR #1269, opened 2020-11-21
Add $conf['sync_chars_regex'] comment about allowed characters

erAck:
Dozens of forum threads resulting from confusion about what is and
what is not (and why not) allowed in sync_chars_regex indicate
there is some explanation needed. Do this in the default config so
that someone actually might read it.

Also see issues #250 #218 #201 #323


plegall:
Good idea to add more explanations. Before merging, why don't you add (in the explanation) some examples of appropriate values for this configuration setting?

erAck:
Done.

PR #1274, opened 2020-12-20
Try to fix HTTPS issues with eg reverse proxies

freemann:


hoedlmoser:
in case if more than one of 

- HTTPS != off or 
- SERVER_PORT == 443 or 
- HTTP_X_FORWARDED_PROTO == https

is true $url will be propagated with multiple 'https://'

e.g. me does
`SetEnvIf X-Forwarded-Proto https HTTPS=on`
due to not every web-app is aware of X-Forwarded-Proto header.

so better concatenate your if-clauses with else.

plegall:

PR #1278, opened 2020-12-31
Show XMP metadata with tools/metadata.php

exekutive:
added ability to read some XMP metadata from images

### DIFF
--- a/tools/metadata.php
+++ b/tools/metadata.php
@@ -79,4 +79,40 @@ $exif = read_exif_data($filename);
 echo '<pre>';
 print_r($exif);
 echo '</pre>';
+
+#
+#        Display XMP metadata using ImageMagick PHP extension
+#
+
+print "<h3>XMP data in '{$filename}'</h3><br />" ;
+print ' (Requires Imagemagick PHP extension)<br />' ;
+print '<pre>' ;
+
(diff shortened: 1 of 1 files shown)


plegall:
Hi @exekutive 

The "problem" is that this pull-request does notgive the ability to Piwigo to use these XMP metadata : `tools/metadata.php` is only a tool, not really connected to Piwigo itself. It's an interesting piece of code showing how to read XMP metadata.

Nowadays, I wonder if it wouldn't be better to rely on a robust tool like `exiftool` to read photo metadata. We wouldn't much care about the difference between EXIF/IPTC/XMP metadata.

exekutive:
@plegall 
The way I understood it, the tools/metadata.php is for diagnostic/debugging purposes.
My PR simply enhances it to read additional useful commonly used metadata
(for example if you were troubleshooting some other XMP function or plugin)

if one wanted to, the method could easily be added to Piwigo in order to capture and "use" the metadata. Just like how other metadata is currently merged into the database.

The problem with exiftool is that it requires shell execution which is often blocked by low cost hosting providers.
It is also more complicated to use. Imagemagick automatically parses and stores the XMP metadata for you. It is much easier to implement.

PR #1282, opened 2021-01-18
A couple of minor PHP 8 fixes that can be applied now

anae-git:
- Use `exif_read_data()` instead of the deprecated `read_exif_data()` alias
- Initialize a couple of uninitialized template variables

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -237,6 +237,13 @@ if ($nb_photos_in_caddie > 0)
       'U_CADDIE' => $link_start.'batch_manager&amp;filter=prefilter-caddie',
       )
     );
+} else {
+  $template->assign(
+    array(
+      'NB_PHOTOS_IN_CADDIE' => 0,
+      'U_CADDIE' => '',
+      )
+    );
 }
 
 // any photos with no md5sum ?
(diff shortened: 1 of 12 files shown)


reubenfarrelly:
This patch applies cleanly but Piwigo still fails when doing an upgrade:

[Mon Feb 01 08:45:01.758069 2021] [php:warn] [pid 14693:tid 140399595222592] [client 2403:5800:7700::] PHP Warning:  Undefined array key "RUN_UPGRADE_URL" in /var/www/piwigo/_data/templates_c/p6jhns^a90b0c1f7aaecbc1badf87b6a648db9d6d3b03bc_0.file.upgrade.tpl.php on line 269

Followed by:

[Mon Feb 01 08:45:13.437938 2021] [php:error] [pid 14693:tid 140398949279296] [client 2403:5800:7700::] PHP Fatal error:  Uncaught Error: Call to undefined function get_magic_quotes_gpc() in /var/www/piwigo/htdocs/admin/include/functions_upgrade.php:215\nStack trace:\n#0 /var/www/piwigo/htdocs/upgrade.php(375): check_upgrade_access_rights()\n#1 {main}\n  thrown in /var/www/piwigo/htdocs/admin/include/functions_upgrade.php on line 215, referer: http://gallery/upgrade.php

reubenfarrelly:
Thanks, that second commit fixes the upgrade process and now it looks like the upgrade is working.  So "works for me" in so far as resolving those magic_quotes_gpc() errors.

Things are still not working with php-8, but based on the errors these are most likely caused by the out of date version of smarty that is bundled with Piwigo.   The latest smarty fixes these problems.  See my issue report #1309 .

klaoun:
hello JFChiken, why not forge your piwigo, then i can test? thanks

Issue #1326, opened 2021-02-06
Code improvement: replace array_push with $arr[] = $new


cypherbits:
Replacing calls to array_push() function with $arr[] = $new is compatible but would improve performance.

PR #1327, opened 2021-02-06
Replace array_push() function call with $arr[] = $add code style. Imp…

cypherbits:
Replace array_push() function call with $arr[] = $add code style. Improves performance. Should not break anything, 100% compatible.

Fixes https://github.com/Piwigo/Piwigo/issues/1326

### DIFF
--- a/admin/cat_move.php
+++ b/admin/cat_move.php
@@ -174,7 +174,7 @@ function assocToOrderedTree($assocT)
     {
       $orderedCat['children'] = assocToOrderedTree($cat['children']);
     }
-    array_push($orderedTree, $orderedCat);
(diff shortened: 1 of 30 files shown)


plegall:
Looks good :-) with some remarks:

- do not modify "vendor" libs : Smarty, phpmailer
- do not modify old migration tasks (install/db/*.php), unless you've tested them all (which I highly doubt). By "old" I mean anything prior to 157
- do not modify install/upgrade_*.php but upgrade_2.10.0.php (the very last one), for the same "ability to test" reason.

cypherbits:
ok, just 4 files now.

Issue #1331, opened 2021-02-06
Loops optimization


cypherbits:
For example: 
`for ($i=0; $i<count($page['chronology_date']); $i++)`

This means on each loop, we call and recount count($page['chronology_date']). 

For performance reasons we could do this:

`$pagecount = count($page['chronology_date']);`
`for ($i=0; $i<$pagecount; $i++)`

plegall:
Is the impact real? I mean if it's to reduce page generation time from 10ms to 9.94ms, it's certainly not where our energy should be focused.

cypherbits:
It depends on the array length and how many loops we change. 
The gain is more than linear, I mean, the more length has the array we are counting, the more performance we gain.

Defining the variable as I suggest is just following best practices without changing any real logic. 

In the search file we changed 13 loops, that is one place where we could feel it.

PR #1332, opened 2021-02-06
Optimize count() in loops

cypherbits:
Fix https://github.com/Piwigo/Piwigo/issues/1331

### DIFF
--- a/admin/cat_move.php
+++ b/admin/cat_move.php
@@ -142,7 +142,7 @@ foreach ($allAlbum as $album)
 
   $parents = explode(',',$album['uppercats']);
   $the_place = &$associatedTree[strval($parents[0])];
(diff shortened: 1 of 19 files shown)


cypherbits:
Could we merge my work?

plegall:
It's not that we don't want any help. It's more that I wonder how important the performance gain is. Because what's sure is that any change might bring regressions. If the risk of regression is higher than the performance improvement, I don't see the point in this specific change.

For example, things like #1380 or #1381 are performance issues more important to solve, in my opinion.

But you're right, we should take more time to merge some PRs. It's just that time is not exactly what we have in excess ;-)

cypherbits:
Hi, I will look into these issues if I can too. This PR will not break anything, just look what it changed...

PR #1376, opened 2021-04-04
Fix URL generation when Piwigo is behind a revers proxy

blastrock:
Since #1274 didn't receive any update in quite some time, here's a new patch to fix the same issue.

I fixed the bug mentioned there by changing less code.


Gerto:
I was having several issues with Piwigo behind a reverse Proxy and can confirm this PR fixed them all.

ZetaKebab:
This solves all my issues, notably the inability to generate image sizes. Should be integrated!

Issue #1325, opened 2021-02-06
international date formats


mistic100:
On the photo property page there is a block with the following text : "Posted the 4 August 2016"
It should be "Posted on August 4th, 2016"

cpol0:
Hi, I'm working on this issue and have a fix.
However, as a new contributor I'm a little lost about the right procedure to do my PR.

1. Is there a recommanded setup for the developper (PHP version, etc...)
2. Is there a continuous integration or unitary test to be sure I will not break everything?
3. If I have general questions about piwigo code, is this issue the best place or is there another place (IRC channel, slack, etc...) to ask?
4. Is there a kind of roadmap where I should look up for most important issues to fix prior others?

For the time being, I have 2 questions:

- is `$user['language']` the best way to get user language?
- to fix this issue, I've used the `IntlDateFormatter` which requires the PHP intl extension. So, it means I'm adding a new dependency to the piwigo project. I wonder if I can decide it myself withour referring to "true" piwigo developers.

Thanks, Pol

plegall:
> However, as a new contributor I'm a little lost about the right procedure to do my PR.

The guide is on https://github.com/Piwigo/Piwigo/blob/master/docs/CONTRIBUTING.md but it may need more details if you didn't find it clear enough :-)

>     1. Is there a recommanded setup for the developper (PHP version, etc...)

Just compatible prerequisites, see https://piwigo.org/guides/install/requirements

>     2. Is there a continuous integration or unitary test to be sure I will not break everything?

No.

>     3. If I have general questions about piwigo code, is this issue the best place or is there another place (IRC channel, slack, etc...) to ask?

For general question, go to https://piwigo.org/forum/viewforum.php?id=24

>     4. Is there a kind of roadmap where I should look up for most important issues to fix prior others?

On this issue tracker (Github), we have set milestones for issues. You can filter on the next milestone if you're looking for the highest priority issues. Next stable milestone is 11.5.0, next unstable milestone is 12.0.0beta1

> For the time being, I have 2 questions:
> 
>     * is `$user['language']` the best way to get user language?

Sure, yes.

>     * to fix this issue, I've used the `IntlDateFormatter` class which requires the PHP intl extension. So, it means I'm adding a new dependency to the piwigo project. I wonder if I can decide it myself withour referring to "true" piwigo developers.

[...]

PR #1384, opened 2021-04-18
issue #1325 - Translation for dates now use IntlDateFormatter

cpol0:
Hi, 
this is my first PR so please kindly tell me if something is not as expected in the PR process.

Thans!


plegall:
Is `IntlDateFormatter` available for any language on a recent version of PHP? We try to avoid having to ask etra modules to install. If you want your Piwigo in Spanish, does the PHP installation needs a "spanish" module to be installed or activated?

In Piwigo, the `{` opens at the beginning of a new line (after indentation), not at the end of the condition line.

The `$lang['Posted the %s']` was actually missing from en_UK, thank you for adding it 👍

Doesn't the `$pattern` needs to be adapted depending on the language? I mean will it output "April 3rd, 2021" in English and "3. April 2021" in German? Because, that's what the current `format_date` doesn't do and what an improvement would do :-)

plegall:
I've written a short script:

```
<?php
date_default_timezone_set('Europe/Paris');
$original = '2021-04-02 14:12:53';

$date = DateTime::createFromFormat('Y-m-d H:i:s', $original);

foreach (array('en_UK', 'fr_FR', 'de_DE') as $language)
{
  $monthname = '';

  echo $language.' : ';

  if ('en_UK' == $language)
  {
    $monthname = 'April';
    printf($date->format('%\s jS, Y'), $monthname);
  }
  elseif ('de_DE' == $language)
  {
    $monthname = 'April';
    printf($date->format('j. %\s Y'), $monthname);
  }
  else
  {
    $monthname = 'Avril';
    printf($date->format('j %\s Y'), $monthname);
  }
  echo "\n";
}
```

and the result is:

```
$ php /Users/plg/www/tmp/date.php 
en_UK : April 2nd, 2021
fr_FR : 2 Avril 2021
de_DE : 2. April 2021
```

... and I would resolve `$monthname` with Piwigo language files.

plegall:
In the code concerning date handling, there are several places where we check for PHP version superior to 5.3. We can consider it's always true in Piwigo 11+. We can even prepare for the near future and say PHP 7.0 will be required for Piwigo 12...

cpol0:
`IntlDateFormatter` is part of the [Intl extension](https://www.php.net/manual/fr/intro.intl.php), so you need to have the Intl extension installed, whatewer the language you set.
I'm a rookie in web development, but it seems this is a very common extension installed by default on a lot of web hosting services. I've checked with hostinger and planethoster, both offers this extension by default, so it **should work by default**. However, I've no idea about the typical config of a classic piwigo server, I think it's better if someone of piwigo team could check this point.

OK for the bracket, I try to code following the [PSR standard](https://www.php-fig.org/psr/psr-12/) but no problem to code with "piwigo style"

No, the pattern is independant from the language. When the pattern is set, it override what you set for `dateType` and `timeType`.
So, even if it doesn't make sense, to keep the backward compatibility in case of there is some code **asking only for the year or the month**, `pattern` is used because there is no `IntlDateFormatter` constant to only display month / year. 
But for common dates format (with or without time), the `IntlDateFormatter` is used without `pattern`. It means it translate according to the right tules of the language (add comma, inverse the place of day / month, etc....). I think this fix the issue, but also fix the issue for all translations available!

I will put some tests results with french, english and german for all availables dates formatsto check with you if the fix is OK.

mistic100:
You can always do a feature detection : if the intl extension is not loaded, use the previous "wrong" date format function.

cpol0:
> You can always do a feature detection : if the intl extension is not loaded, use the previous "wrong" date format function.

Smart! I didn't know that we can do that in PHP, thanks for the advice! I will look at this point.
On my side, I've checked several locale on my `format_date `function.
Instead of use `$user['language']` variable in `datefmt_create` function, I've force manually some locales for tests. 
In the results bellow:

- the first string is the post date of a picture (available in the admin panel)
- the second is the date of a comment on the picture
- the third is a date that I've forced manually with the following code `$show = array('month', 'year');` to simulate a "strange" date format with only a month and a year. Here are my results:

fr_FR (cocorico)
Samedi 17 Avril 2021
Dimanche 18 Avril 2021 11:44
avril 2021

en_US (oncle sam)
Saturday, April 17, 2021
Sunday, April 18, 2021 at 11:44 AM
April 2021

de_DE (german)
Samstag, 17. April 2021
Sonntag, 18. April 2021 um 11:44
April 2021 (yes, April is a german month, surprising no?)

es_ES (spanish)
sábado, 17 de abril de 2021
domingo, 18 de abril de 2021, 11:44
abril 2021

zh_CN (china simplified, not traditionnal)
2021年4月17日星期六
2021年4月18日星期日 上午11:44
四月 2021

ru_RU (russian)
суббота, 17 апреля 2021 г.
воскресенье, 18 апреля 2021 г., 11:44
апреля 2021

In my docker setup, there is no locale installed manually (I guess the default is probably en_US....).

I will look at the @mistic100 suggestion, but if we can fallback on the old formatting style if the intl extension is not installed, I think the fix is acceptable.

cpol0:
Hmmmm I force the push with -f option, I wonder if it was a good idea but I didn't want to push a 2nd commint over the 1st one. What should I do when I rewrite my PR? 

By the way, you can review the new PR:
I do a check extension to switch between the legacy format and the intl format.
I did a short test with intl extension installed and Russian harcoded in my `datefmt_create` function. The result for a picture date is: `суббота, 17 апреля 2021 г.`
Then I remove the intl extension from my docker setup and as a guest visitor (default is english, the result is the one displayed by the legacy format: `Saturday 17 April 2021`

Pol

Issue #184, opened 2015-07-25
[photos] Zero month bug


plegall:
**Reported by eric on 20 Feb 2012 21:31**

**Version:** 2.3.3

In batch manager, we can change the creation date for each photos and we can choose to enter only the day or the month or the year or all of them. But if choose to skip the month like this : 12 ---- 1970, a php notice appears on picture.php for the related photo:

Notice: Undefined offset: 0 in ..\include\functions.inc.php on line 526.

**Additional information:** Bug reported by wernerk with a possible fix: http://piwigo.org/forum/viewtopic.php?id=18881

[Piwigo Bugtracker #2582](http://piwigo.org/bugs/view.php?id=2582)

PR #1418, opened 2021-05-25
Check the X-Forwarded-Proto header for HTTPS

bobobo1618:
Should fix issues like https://github.com/Piwigo/Piwigo-Android/issues/184 all proper-like.

Currently, Piwigo only uses a HTTPS URL scheme if PHP's `$_SERVER['HTTPS']` variable returns true but this usually isn't the case where a reverse proxy is involved. In those cases, the server can set `X-Forwarded-Proto`.

### DIFF
--- a/include/functions_url.inc.php
+++ b/include/functions_url.inc.php
@@ -37,8 +37,11 @@ function get_absolute_root_url($with_scheme=true)
   if ($with_scheme)
   {
     $is_https = false;
-    if (isset($_SERVER['HTTPS']) &&
-      ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
+    $server_https = isset($_SERVER['HTTPS']) &&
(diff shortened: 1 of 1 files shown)


ioogithub:
> Currently, Piwigo only uses a HTTPS URL scheme if PHP's `$_SERVER['HTTPS']` variable returns true but this usually isn't the case where a reverse proxy is involved. In those cases, the server can set `X-Forwarded-Proto`

Can you explain more what you mean by the server can set  X-Forwarded-Proto for users who come across this issue? Where is that set? I am still having to use the hack as outlined in Issue #184 to get images in the Android app.

ioogithub:
Oh wait, has this fix been merged?

bobobo1618:
The fix hasn't been merged, this PR is still open. If the fix is merged, you'll need to set the header to "https" rather than "http".

plegall:
An easy and immediate solution is given on https://github.com/Piwigo/Piwigo/pull/1637#issuecomment-1166576255

bobobo1618:
Since https://github.com/Piwigo/Piwigo/pull/483 is around, I'll close this PR.

Issue #1363, opened 2021-03-10
Deprecated: Invalid characters passed for attempted conversion, these have been ignored in include/template.class.php:1024


giga:
Hello,

I recently got this error (this happened with php 7.4 afaik).

Deprecated: Invalid characters passed for attempted conversion, these have been ignored in include/template.class.php:1024

This is because crc32 can return negative integers. According to php documentation, this happens a lot with 32 bit systems, and sometimes with 64 bit systems.

I changed the `base_convert` call to `base_convert(hash("crc32b", $compile_id), 16, 36)`, and this fixes the problem. Unless I'm mistaken this should be equivalent to the current code.

Sorry I'm too lazy too submit a PR for a one-liner.

giga:
There are other calls to crc32 which should be fixed as well.

rmiddle:
Will including this cause an issue with 64 Bit OS's?

faginbagin:
I'm getting four deprecated messages at the top of the web page when I first go to my piwigo site:

- Deprecated: Invalid characters passed for attempted conversion, these have been ignored in .../piwigo/include/template.class.php on line 215
- Deprecated: Invalid characters passed for attempted conversion, these have been ignored in .../piwigo/include/template.class.php on line 1024
- Deprecated: Invalid characters passed for attempted conversion, these have been ignored in .../piwigo/include/template.class.php on line 1024
- Deprecated: Invalid characters passed for attempted conversion, these have been ignored in .../piwigo/include/template.class.php on line 1899

I'm in the process of upgrading from debian stretch to bullseye on a 32 bit pentium 4 machine. I've got both debian versions installed in a multiboot config. I haven't upgraded to piwigo 12 yet. From what I can tell looking at the source, commits and this issue, these haven't been fixed in 12. I'll go back to stretch and upgrade to piwigo 12, WHEN/IF the plugins I use are supported on 12. Then I'll periodically check bullseye when there are new piwigo releases. But, if there are patches to test, I'd like to help. Ideally (for me) they would be backward compatible with debian stretch because I'm using a partition shared by my bullseye and stretch installs for piwigo.

alexmogavero:
I am having this same issue on the raspbian bullseye, with raspberry pi 4.
@ludo-c , since you added a fix on your fork, would it be possible to issue a pull request here?

ludo-c:
@alexmogavero Hello,

A PR is already opened : https://github.com/Piwigo/Piwigo/pull/1420

[...]

PR #1420, opened 2021-05-30
issue #1363: Fix crc32 for 32bit processors

ludo-c:
Hello,

I got the issue on a Raspberry Pi 3.

Thanks to @giga for the fix, here is the corresponding PR.

### DIFF
--- a/admin/include/pclzip.lib.php
+++ b/admin/include/pclzip.lib.php
@@ -2677,7 +2677,7 @@
         @fclose($v_file);
 
         // ----- Calculate the CRC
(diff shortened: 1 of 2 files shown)


chancsy (commented):
Confirming this solving problems after first time installation on Synology DS218j which runs on 32-bit processor. Without this fix plugins cannot be activated, photo upload will stuck at 100% and become orphan, unable to operate album dropdown list when uploading photo, unable to see user list under Users->Manage and various other issues making Piwigo total unusable.

shapirus:
Note that [3239296] contains only numbers. Must be a lucky commit!

Issue #911, opened 2018-08-10
No mobile history log when mobile theme not default


7tonin:
If your default theme is "elegant", and a mobile device browse the gallery, there is no history log for this device, despite the use of "smart pocket" mobile theme.

The solution may be in the answer of "How can we get run add_event_handler('ws_add_methods', 'sp_add_methods'); from themes/smartpocket/themeconf.inc.php **_earlier_** ?"

PR #1421, opened 2021-05-31
Issue #911: bringing missing methods

7tonin:
SmartPocket log history is missing some methods. Looks like this copy/paste hack solves the issue. I cannot really explain why, sorry.

This may help find an idea to improve the code and to cleanly solve the issue.

### DIFF
--- a/ws.php
+++ b/ws.php
@@ -19,6 +19,65 @@ if ( !$conf['allow_web_services'] )
 
 include_once(PHPWG_ROOT_PATH.'include/ws_core.inc.php');
 
+/* HACK BEGIN */
+// copy/paste from themes/smartpocket/themeconf.inc.php
+add_event_handler('loc_end_section_init', 'sp_end_section_init');
+function sp_end_section_init()
+{
+  global $page, $template;
+
+  // variables to log history
+  $template->assign(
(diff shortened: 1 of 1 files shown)


7tonin:
#911

plegall:
I'm reviewing this issue/pull-request. The problem is real (`smartpocket.images.logHistory ... Method name is not valid`) but this way to solve the problem is not the right way.

The technical problem is that the `themes/smartpocket/themeconf.inc.php` is not loaded when running `ws.php`. Unfortunately, this is where the method `smartpocket.images.logHistory` is defined.

I propose to create a new core API method `pwg.history.log` which would do the exact same as the specific `smartpocket.images.logHistory` but any theme/plugin would be able to use it without caring about being loaded by the theme/plugin.

Once this API method is implemented in Piwigo core, we can remove `smartpocket.images.logHistory` from SmartPocket and use the new core API method instead.

@7tonin do you feel ready to follow this plan?

plegall:
I close this pull-request because I made it "another way". As described in my previous post, I've implemented a new core API method `pwg.history.log`. See 002abf87df0ed58183ab6f794b9b620cbe9cf083

PR #1427, opened 2021-06-11
Fix missing thumbnail for TIFF images.

weakset:
Without this, the thumbnail is replaced with unknown MIME icon.


flop25:
Thank you
quick and useful fix

PR #1431, opened 2021-06-17
delete comments fixIE

klaoun:
hello everyone
I suggest you delete the comments because we are in the future ie7. What do you think ? thanks


klaoun:
hello everyone

PR #1435, opened 2021-06-18
Convert uploaded TIFF and delete the original

weakset:
1. Allow resizing TIFF 
2. Delete uploaded TIFF after the representative is made.
3. Register the representative as new "original".

Web browsers cannot display TIFF, so it is waste of space to keep them. This patch does only apply to new uploads.


weakset:
I found some issues with default mobile theme and also center of interest didn't seem to update. It is unclear to me which issues are bugs in Piwigo and which are bugs in themes, so I'll keep further hacks in my personal fork. I still think these modifications are useful, but I have no time to trace every compatibility issues.

PR #1575, opened 2021-12-04
PHP 8: check for type object before calling get_class()

erAck:
Otherwise get_class() throws fatal error.
Reported in https://piwigo.org/forum/viewtopic.php?id=31711


erAck:
@plegall Not interested in merging this?

erAck:
@MatthieuLP interested in merging this?

nitro322:
Thanks, @erAck.  I stumbled across this while trying to figure out why I could no longer use digiKam to export photos to piwigo - REST authentication is broken on PHP 8.x without this patch.  Working great with.  Much appreciated.

plegall:
What about using `if ($response instanceof PwgError)` instead?

erAck:
Sure, that's better, done. Found additional two occurrences.

PR #944, opened 2018-11-22
pwg.categories.getImages improvements

ethiry:
This pull request is about 3 small improvements for pwg.categories.getImages that are useful in my use :
- add latitude and longitude
- decrease the payload of the reply by making optional data I don't need : derivatives and categories

PR #1579, opened 2021-12-09
Closes #944: [WebAPI] Add 'display' parameter to categories.getImages

exastion:
This pull request implements a `display` parameter for the API call `pwg.categories.getImages` similar to `pwg.users.getList`.

It is similar to #944 but goes further:
- Makes the changes proposed there, that were never implemented
- Exposes additional fields

As there was no reaction on that pull request, I decided to create my own with the proposed changes.

Closes #944


plegall:
I'm reviewing this pull-request (just letting you know)

PR #1580, opened 2021-12-11
Add last calculated cache size to Dashboard Storage chart

erAck:
Size of cached files with all derivates and representatives may be
as large as all image files, actually doubling the needed storage
size. So it may be nice to see it represented in the storage
chart, if one remembers to occasionally recalculate it under
Tools -> Maintenance..

To not have this mis-representation active for non-aware users,
make it a configuration switch defaulting to false.


plegall:
I like this BUT, could you switch the configuration setting to `true` by default. It's nice to be able to deactivate it (and we will deactivate it on Piwigo.com) but it should be active by default, in my opinion.

PR #1585, opened 2021-12-20
Add option to turn off display_errors

balki:
This option can be used if we want the php errors/warnings in the server logs but not on the browser


erAck:
See https://piwigo.org/forum/viewtopic.php?pid=178575#p178575

erAck:
See https://piwigo.org/forum/viewtopic.php?pid=180778#p180778

PR #1597, opened 2022-01-06
Modifying image-class to improve the usage of external image converte…

thefrip:
Changing the image class to make it more efficient when using external image converter:
- adding a global variable ext_imagick_version to set the imagick version instead of forking an exec each time which is quite inefficient
- relying on internal PHP function getimagesize to get the image size instead of forking
- clarifying the return code from add_command (

### DIFF
--- a/admin/include/image.class.php
+++ b/admin/include/image.class.php
@@ -317,6 +317,11 @@ class pwg_image
   {
     global $conf;
 
+    if (isset($conf['ext_imagick_version'])) {
+        self::$ext_imagick_version = $conf['ext_imagick_version'];
+        return true;
+    }
+
     if (!function_exists('exec'))
     {
       return false;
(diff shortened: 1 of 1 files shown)


thefrip:
Hi @plegall

ext_imagick_version is not set by default, it is an option to be set manually in the Local Configuration File for those who want to use it. If not set, then normal behaviour applies. This is why the "reset" would happen in the Local Configuration File and nowhere else.

And regarding using the convert command instead of the PHP function makes it super inefficient because of multiple forks happening when for instance viewing an album for the first time, when thumbnails have to be created. With the changes I propose, loading time decreased drastically.

plegall:
> ext_imagick_version is not set by default, it is an option to be set manually

I hadn't understood. It's not written anywhere ;-) So if I add `$conf['ext_imagick_version'] = '7.1.2';` in `local/config/config.inc.php` the result would be that function `is_ext_imagick` will no longer "really try" if ImageMagick command line is available and simply rely on the information provided in the configuration by the user.

It's a "dangerous" setting in my opinion, but I understand it avoids a useless call to the `exec` function (what you call a "fork").

In my opinion f8ba24a7776cc746d2a75238040c73ffae199376 is no longer useful but reverting it (getting rid of `$ext_imagick_version`) does not mean removing the test of the command-line to make sure `convert` is available. The configuration setting could be more something like `$conf['ext_imagick_check_cli_before_use'] = true;` (by default).

> And regarding using the convert command instead of the PHP function

Don't you think it was implemented this way to make sure ImageMagick was going to be able to process (resize) the file afterwards? I would like @modus75 to give his opinion about that.

thefrip:
Hi @plegall

> Don't you think it was implemented this way to make sure ImageMagick was going to be able to process (resize) the file afterwards? I would like @modus75 to give his opinion about that.

I haven't seen any answer from @modus75 but to be honest I am not sure to understand why the use of `convert` at the moment of getting the image size would "guarantee" anything afterwards (but I may be missing the point, happy to better understand the rationale). The thing is that any call to anything external to PHP (`convert` in that case) will create an `exec` which will be detrimental to the overall performance. So, if there is an equivalent built-in function, I'd rather use it in the name of performance if you see what I mean.

I hope my explanations clarify the pull request :-)

PR #1608, opened 2022-02-08
Fix dutch translation typo in admin.lang.php

barrydegraaff:
Fix wrong translation in admin UI -> manage album. The field name for the album should be Naam (Name) and not the 'photographers title/source'.


ddtddt:
Hi :-)

Is better register on translate tool -> https://piwigo.org/translate/
Thanks

PR #1615, opened 2022-02-25
Include favorite status in image list API calls

abrenoch:
Will include an `is_favorite` property in the `pwg.categories.getImages` and `pwg.tags.getImages` API methods (if the user is logged in)


abrenoch:
Hey @plegall that does seem like a much nicer solution! I'll close this PR and open a new one with your proposed changes.

Thanks for reviewing!

plegall:
no need to close this one, simply make some new commits. It will update the pull-request :-)

abrenoch:
okay - was thinking of just keeping the commit history a little cleaner. But if that is okay with you it certainly is with me 👍 .

I'll update this in the next day or 2!

plegall:
> okay - was thinking of just keeping the commit history a little cleaner.

We have to deal with 2 histories :

1. the commit history : don't worry, it's not because you do 3 commits that we won't "squash" them before merge ;-)
2. the issues/pull-requests history : this one is harder to squash, that's why I prefer that you stay on the same pull-request. It will be much simpler to understand on the long run.

abrenoch:
Revised per your advice! also added the favorite checking bit to the image search function

plegall:
OK, now you do the exact same on 3 different places, let's create a dedicated function `get_user_favorites` in file `include/functions_user.inc.php` right after function `check_user_favorites` :

```
/**
 * List favorite image_ids of the current user.
 * @since 13
 */
function get_user_favorites()
{
  global $user;

  if (is_a_guest())
  {
    return array();
  }

  $query = '
SELECT
    image_id,
    1 as fake_value
  FROM '.FAVORITES_TABLE.'
  WHERE user_id = '.$user['id'].'
';

  return query2array($query, 'image_id', 'fake_value');
}
```

(I don't write `[]` but `array()` on purpose)
(I open `{` on a new line, on purpose)

Then in your 3 functions, simply write:

```
$favorite_ids = get_user_favorites();
...
$image['is_favorite'] = isset($favorite_ids[ $row['id'] ]); // nicer than my 4 lines of code
```

(I know I ask some changes I could do myself afterwards, but I think it's better on the long run if I give this kind of tips)

abrenoch:
I should have known better :p
Thanks for the tip on where to place that - looks much better!

PR #1623, opened 2022-03-08
Make get_fs_directories non recursive

kot0dama:
This will make get_fs_directories non recursive, thus fixing a potential stack overflow when operating on directories with quite large depth. It might change a bit the order directories appear in the list though.

Feel free to amend any way you like.

Thank you,
Loïc

### DIFF
--- a/admin/include/functions.php
+++ b/admin/include/functions.php
@@ -605,20 +605,25 @@ function get_fs_directories($path, $recursive = true)
 
   if (is_dir($path))
   {
-    if ($contents = opendir($path))
+    $remaining = array($path);
+    $dirs[] = $path;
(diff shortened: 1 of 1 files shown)


plegall:
Hi @kot0dama 

I need to make more tests but you may be able to answer this questions already:

> thus fixing a potential stack overflow

Can you give an example of when (in depth of filesystem) this stack overflow might happen? I've never seen any so I wonder.

> It might change a bit the order directories appear in the list though

This might be a problem. In the synchronization process, it's important that we register `dir1/dir1.1` after `dir1`.

Now concerning speed: did you make some speed checks? How does it behaves compared to the recursive version of `get_fs_directories`.

PR #1630, opened 2022-03-15
Updated swedish translation.

pbobbenb:
Made an update to the Swedish translation for Piwigo.


ddtddt:
Sorry for delays. For translation can you register and use translate tool -> https://piwigo.org/translate
when you register, i open right for you

Issue #1527, opened 2021-10-18
[Bug] Piwigo change HTTPS port to 80 due to bad awareness of reverse proxy


LesCarnets:
Hello,

Each time I am doing upgrades, I breack my Piwigo (deployed on shared hosting managed services).

# Bug
Several picies of codes are doing calls where port 80 is forced in the URL even if we are in HTTPS (for example, authentication is brocken).

# Resolution
I can see on forums this is a recurent issues and a lot of people are facing this issue.

It seems the correction is already provided by the community:
https://fr.piwigo.org/forum/viewtopic.php?id=28052#p225699 (sorry, this forum is in French)

On my side, I confirm this correction is well working.

## Environment
- **Version**: 11.5.0
- **OS**: Linux webdb46.lwspanel.com 4.19.0-6-amd64 # 1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64
- **Web server**: Linux shared hosting
- **PHP**: 7.4.23
- **MySQL**: 5.5.5-10.3.25-MariaDB-0+deb10u1
- **ImageMagick**: 6.9.7-4

Thanks and best regards,

PR #1637, opened 2022-03-24
use HTTP_X_FORWARDED_PROTO when behind a proxy

blink38:
When behind a proxy, absolute root url is construct using HTTP_X_FORWARDED_HOST but is ignoring the scheme.

The modification proposed is to use HTTP_X_FORWARDED_PROTO which include the scheme when construct the absolute root url.

Use case : 
Piwigo is running in docker container which expose the service with HTTP. A traefik proxy is exposing the service via HTTPs. Without this modification, url are return with http:// scheme.

May resolve issue  #1527

### DIFF
--- a/include/functions_url.inc.php
+++ b/include/functions_url.inc.php
@@ -37,8 +37,11 @@ function get_absolute_root_url($with_scheme=true)
   if ($with_scheme)
   {
     $is_https = false;
-    if (isset($_SERVER['HTTPS']) &&
-      ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
+    if ((isset($_SERVER['HTTPS']) &&
+        ((strtolower($_SERVER['HTTPS']) == 'on') or ($_SERVER['HTTPS'] == 1)))
+        ||
+        ((isset($_SERVER['HTTP_X_FORWARDED_PROTO']) &&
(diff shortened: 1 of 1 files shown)


gasparakos:
It is working for Us!

Our system runs in docker, behind an apache reverse proxy. Container port 80 is exposed on 12380. On apache `RequestHeader set X-Forwarded-Proto 'https'` was set but some functionality didn't worked, eg: file upload or mobile client show only white square insted of pictures. After apply 4eb8765618f4934dffcf7f0767c77c29c85c379a problems was resolved.

Thank You!

mossroy:
For information, I had created an almost identical PR 6 years ago: #483 (that was not merged either, unfortunately)

To me, this is still relevant, and I still workaround it with the following code in `local/config/config.inc.php`:
```
if ($_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https'){
       $_SERVER['HTTPS']='on';
}
```

plegall:

blink38:
The proposed code by @mossroy https://github.com/Piwigo/Piwigo/pull/1637#issuecomment-1166576255 is a workaround, waiting for the MR (#483 or #1637) to be accepted. 

It's not relevant to say : the solution is to put the fix code in the config.php instead of correcting the code where it must be done.

PR #1681, opened 2022-07-02
Show the current version on the Updates page.

helmo:
After receiving an update notification I was looking for the current version number.
However I did not find that on the updates page.  This patch adds it to the bottom of that page.
It also fixes a missing space in the Dutch translation.

image


plegall:
I'm going to make some slight presentation changes but I merge this pull-request : it's a good idea. Thank you @helmo

PR #1633, opened 2022-03-19
Include PHPMailer Exception in pwg_mail()

erAck:
Otherwise if an exception is to be thrown it results in
Fatal error: Uncaught Error: Class 'PHPMailer\PHPMailer\Exception' not found
See https://piwigo.org/forum/viewtopic.php?id=31942

PR #1682, opened 2022-07-03
Include PHPMailer Exception class on SMTP error

joshumax:
As it currently stands, any issue sending mail over SMTP using PHPMailer causes a PHP exception: "Uncaught Error: Class 'PHPMailer\PHPMailer\Exception' not found". This commit includes the PHPMailer\Exception class to prevent this issue and also provide useful connection information to the Piwigo admin.


plegall:
I'm going to merge the same pull-request #1633 by erAck

PR #1688, opened 2022-07-11
qsearch_results should change the ids of the result set, otherwise th…

### LINKED ISSUE #1686: impossible to change search_results during handling of event "qsearch_results"
I'm currently trying to extend quicksearch to search contents of private fields. For this i use the eventHandler 
"qsearch_results" wich allows me to add the so found items to $search_results['items'].
 This works, as long as there is no match on the core fields. If there is such a match, then the content of $search_results['items'] wil be overwritten by later code during permission checks.
IMO the handler for qsearch_results should receive a copy of the $ids array, an return a new $ids array (changed or not)
$ids = trigger_change('qsearch_results', $ids, $expression, $qsr);

tkempf:
…e changed_results will be overwritten by the following code during permission checks fixes #1686

### DIFF
--- a/include/functions_search.inc.php
+++ b/include/functions_search.inc.php
@@ -1428,9 +1428,9 @@ function get_quick_search_results_no_cache($q, $options)
   $search_results['qs']['matching_tags'] = $qsr->all_tags;
   $search_results['qs']['matching_cats'] = $qsr->all_cats;
   $search_results = trigger_change('qsearch_results', $search_results, $expression, $qsr);
-  if(isset($search_results['items']))
+  if (isset($search_results['items']))
   {
-    $ids=array_merge($ids,$search_results['items']);
+    $ids = array_merge($ids, $search_results['items']);
   }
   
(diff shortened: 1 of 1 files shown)


plegall:
In trigger `qsearch_results` you're replaces `$search_results` (which is a multiple-level associative array) with `$ids` (which is simple array). It might have an impact on code using this trigger.

See for example https://github.com/Piwigo/piwigo-gmaps/blob/master/qsearch.inc.php#L227 : in plugin RV Maps & Earth, function `rvm_qsearch_results` is expecting `$sr` (mapped on `$search_results` variable) to be this multiple-level array.

Shouldn't we have the `$ids` replaced by `$search_results['items']` instead ? (and skip line `$search_results['items'] = $ids;` at the end of function `get_quick_search_results_no_cache`)

plegall:
Looks much better to me.

Small last step, can you add a space:

-  after `if`
- before and after `=`
- after `,`

PR #1702, opened 2022-08-03
Correct use_iptc_mapping, image author is '2#080' not '2#122'

erAck:
Also comment out date_creation iptc_Date_Created 2#055 mapping as
it overrides EXIF DateTime with _only_ date, IPTC has an extra
field iptc_Time_Created 2#060 and both would have to be combined.

See https://piwigo.org/forum/viewtopic.php?pid=184200#p184200 and https://piwigo.org/forum/viewtopic.php?pid=168381#p168381

### DIFF
--- a/include/config_default.inc.php
+++ b/include/config_default.inc.php
@@ -339,8 +339,10 @@ $conf['use_iptc'] = false;
 // associates a piwigo_images column name to a IPTC key
 $conf['use_iptc_mapping'] = array(
   'keywords'        => '2#025',
-  'date_creation'   => '2#055',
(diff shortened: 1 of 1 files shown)


plegall:
The problem with these mappings is that all software don't use the same IPTC tag for the same information.

I don't want to change the default setting. It's the one we have had for a very long time and changing it might break many existing installation of Piwigo. Instead we could have a wiki page, with recommendation of `$conf['use_iptc_mapping']` depending on the software used to enter IPTC.

erAck:
Hum.. ok, closing.

PR #1708, opened 2022-08-20
Fix #1692 Remove Content-Length header on images (HTTP2 compat)

### LINKED ISSUE #1692: Error loading images over HTTP2
I updated Chrome today which seems to have switched to HTTP2 by default. When viewing a photo via `i.php` the image is broken and there is an error in the console :
```
GET https://galerie.strangeplanet.fr/i.php?/upload/2022/07/16/20220716130608-9f2835c1-sq.jpg net::ERR_HTTP2_PROTOCOL_ERROR 200
```

You can test it here by clicking on any blue pin https://galerie.strangeplanet.fr/osmmap.php?/categories&v=2

Or here https://galerie.strangeplanet.fr/picture.php?/6993/category/231 by switching to the XS resolution (assuming the image has not been generated and cached, so it will only be broken once per image/resolution).

I did some search and found this discussion : https://stackoverflow.com/questions/58215104/whats-the-neterr-http2-protocol-error-about  
It _may_ be related to the `Content-Length` header which is more striclty checked in HTTP2, and the size is obviously wrong when gzip is enabled on the server.

mistic100:
Content-Length header should be the body size after compression (if enabled) and not the size of the original body. In HTTP1.1 such invalid value is not problematic but in HTTP2 it prevents images to display.

**WARNING** : this must be tested on a HTTP1.1 server before shipping

### DIFF
--- a/i.php
+++ b/i.php
@@ -349,7 +349,6 @@ function send_derivative($expires)
   {
     header('Expires: '.gmdate('D, d M Y H:i:s', $expires).' GMT');
   }
-  header('Content-length: '.$fstat['size']);
   header('Connection: close');
 
   $ctype="application/octet-stream";


plegall:
Seems fine on HTTP 1.1

Issue #1715, opened 2022-09-06
Can't update extensions in 13.0.0RC3


rootflags:
Was getting the following error when trying to update my extensions:

`[Tue Sep 06 11:33:23.152166 2022] [php:info] [pid 2466] [client 192.168.1.250:64629] PHP Deprecated:  Implicit conversion from float 1662482003.152163 to int loses precision in /data/photos/software/piwigo/include/Logger.class.php on line 367
`

I patched this on my system, fix below:

```
--- piwigo/include/Logger.class.php.orig        2022-09-06 11:35:08.660271896 -0500
+++ piwigo/include/Logger.class.php     2022-09-06 11:34:18.356824710 -0500
@@ -364,7 +364,7 @@
   {
     $originalTime = microtime(true);
     $micro = sprintf('%06d', ($originalTime - floor($originalTime)) * 1000000);
-    $date = new DateTime(date('Y-m-d H:i:s.'.$micro, $originalTime));
+    $date = new DateTime(date('Y-m-d H:i:s.'.$micro, intval($originalTime)));
     return $date->format($this->options['dateFormat']);
   }

```

PR #1716, opened 2022-09-06
Update Logger.class.php to fix bug #1715

rootflags:
13.0.0RC3 plugin updates fail with "Implicit conversion from float to int loses precision".  This fixes the issue (like 367 - simply forces $originalTime to an integer).

### DIFF
--- a/include/Logger.class.php
+++ b/include/Logger.class.php
@@ -364,7 +364,7 @@ class Logger
   {
     $originalTime = microtime(true);
     $micro = sprintf('%06d', ($originalTime - floor($originalTime)) * 1000000);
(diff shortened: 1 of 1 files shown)


plegall:
Hi @rootflags 

I'm OK for this:

```
-    $date = new DateTime(date('Y-m-d H:i:s.'.$micro, $originalTime));
+    $date = new DateTime(date('Y-m-d H:i:s.'.$micro, intval($originalTime)));
```

but not for changing the whole file, as what the "Files changed" tab shows us in the pull-request. Can you update your pull-request or should I simply apply the real change?

plegall:
OK, done in https://github.com/Piwigo/Piwigo/commit/a2da870d3abce1bd0ff02da249c3c09ffb4b8ffb

PR #1749, opened 2022-10-12
Move sending headers before the flush, fixes #1746

### LINKED ISSUE #1746: Warning: Cannot modify header information - headers already sent in /var/www/piwigo/action.php on line 228
I'm getting the above warning when trying to download an image in Piwigo 13.0.0. This corrupts the download.

Could be related to dd44d0e40999632fb1d80f313f51b6decbfa937c, as I think headers can't be sent after calling `flush()`

erAck:
There was
Warning: Cannot modify header information - headers already sent in action.php on line 228 see https://piwigo.org/forum/viewtopic.php?id=32254

Unclear how headers could had been sent before sending headers, but..

### DIFF
--- a/action.php
+++ b/action.php
@@ -216,17 +216,17 @@ if (ini_get('safe_mode') == 0)
   @set_time_limit(0);
 }
 
+foreach ($http_headers as $header)
+{
+  header( $header );
+}
+
 // Without clean and flush there may be some image download problems, or image can be corrupted after download
 if (ob_get_length() !== FALSE)
 {
   ob_flush();
 }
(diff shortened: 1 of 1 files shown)


plegall:
@MatthieuLP can you test this PR with both PHP 7 and 8 please?

plegall:
tested on PHP 8.1. Working fine.

plegall:
oups, was merged on 13.x while we should always fix bugs on master and then we cherry-pick it on 13.x :-)

This time I'm going to cherry-pick on master instead.

Issue #1753, opened 2022-10-16
syntax error, unexpected ')' in include/ws_functions/pwg.php caused by trailing commas


drazisil:
https://github.com/Piwigo/Piwigo/blob/ba08bffcb2768fb773c27d9b7ddb0f1d7645750b/include/ws_functions/pwg.php#L832

and 

https://github.com/Piwigo/Piwigo/blob/ba08bffcb2768fb773c27d9b7ddb0f1d7645750b/include/ws_functions/pwg.php#L838 

have an invalid trailing comma.

They generate the following error when trying to connect with the remote sync client

```
2022/10/16 15:50:49 [error] 1167#1167: *320 FastCGI sent in stderr: "PHP message: PHP Parse error:  syntax error, unexpected ')' in /var/www/***/include/ws_functions/pwg.php on line 833" while reading response header from upstream, client: 172.70.110.35, server:*** request: "POST /ws.php HTTP/1.1", upstream: "fastcgi://unix:/run/php/php7.2-fpm.sock:", host: "***"
```

PR #1754, opened 2022-10-16
issue 1753 : fix syntax error API

Thibautg16:
syntax error, unexpected ')' in include/ws_functions/pwg.php
fix issue #1753 
piwigo 13


Thibautg16:
fix issue #1753

PR #1785, opened 2022-11-13
Support webp format

Phlogi:
This allows users to easily add further formats.

Also these formats should not be hard-coded there anyway :) 

Update: multiple changes now to support serving webp files.

### DIFF
--- a/action.php
+++ b/action.php
@@ -21,6 +21,7 @@ function guess_mime_type($ext)
     case "jpg": $ctype="image/jpeg"; break;
     case "png": $ctype="image/png"; break;
     case "gif": $ctype="image/gif"; break;
+    case "webp": $ctype="image/webp"; break;
     case "tiff":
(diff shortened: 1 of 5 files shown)


NeiRib:
Thumbnails don't work

edison23:
Hi, works great for me (used the steps described here - https://edison23.net/blog/posts/piwigo-webp-support#implement-the-pr-into-your-piwigo ), but I'm missing EXIF support. 

Were you, @Phlogi , able to look into that, if it's possible to solve somehow? Because currently, even though Piwigo does display the WebP images without an issue, it sorts them wrong because it can't see the EXIF tags, so it's rather unusable :(

plegall:
> [...] even though Piwigo does display the WebP images without an issue, it sorts them
> wrong because it can't see the EXIF tags, so it's rather unusable :(

I will work on it (after this pull-request is merged), and other things related to webP.

Issue #1107, opened 2019-10-25
Response in text/plain instead of application/json


sridharb1:
If I query the webservice with format=json, I expect the response to be in application/json. It is in text/plain. Could it be corrected?

Latest piwigo, php 5.6.40.

EddyLB:
In fact, most answers are in JSON format but the server returns plain text messages when it encounters errors, which is problematic for third party apps.

EddyLB:
I have replaced `text/plain` with `application/json` in the file `/include/ws_protocols/json_encoder.php` and everything works fine with the iOS app awaiting for JSON data. However, web browsers have issues and loading the `tools/ws.htm` page returns an 'error when parsing JSON string'. So some work needs to be done to fix this issue before switching to 'application/json'.

Issue #448, opened 2016-04-03
pwg.images.add returns a plain text error, regardless of requested format


vvinogradsky:
An attempt to upload an unsupported file format produces a plain text error, even though JSON is explicitly was requested.

```
REQUEST:

POST http://piwigo28.int/ws.php?method=pwg.images.add&format=json HTTP/1.1
Accept: */*
Cookie: pwg_id=1p64dj3gpn9jap5j03lk59f2e3
Content-Type: application/x-www-form-urlencoded
Content-Length: 261
User-Agent: Adobe Photoshop Lightroom
Host: piwigo28.int
Pragma: no-cache

categories=2&file_sum=dbdae213bd98acaade001d2cb582827d&comment=&author=&tag_ids=1&pwg_token=4f9a19ce4c47487f45e65d816c715553&name=&original_sum=dbdae213bd98acaade001d2cb582827d&date_creation=2014-09-01%2022%3A49%3A37&level=0&resize=0&original_filename=-9035.mp4

RESPONSE:

HTTP/1.1 200 OK
Date: Sun, 03 Apr 2016 17:49:28 GMT
Server: Apache/2.2.22 (Win32) mod_ssl/2.2.22 OpenSSL/0.9.8t PHP/5.3.29
X-Powered-By: PHP/5.3.29
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
Pragma: no-cache
Content-Length: 19
Content-Type: text/html; charset=utf-8

forbidden file type
```

Tested with Piwigo 2.5 and 2.8.0.

PR #1848, opened 2023-01-07
Fix mediatype for api json format

### LINKED ISSUE #1847: Piwigo API returns wrong mediatype
When using format=json on the api, the wrong mediatype is returned.
This causes a problem when using different frameworks in java since automatic decoding is not supported.

For example the following code snippet will return an exception:

```
ClientConfig clientConfig = new ClientConfig().register(GsonMessageBodyHandler.class);
Client client = ClientBuilder.newClient(clientConfig);
Response response = client.target("https://<hostname>")
   .path(String.format("ws.php"))
   .queryParam(FORMAT_JSON.getKey(), FORMAT_JSON.getValue())
   .queryParam(IMAGE_INFO.getKey(), IMAGE_INFO.getValue())
   .queryParam(IMAGE_ID, imageId)
   .request(MediaType.APPLICATION_JSON).get();
return response.readEntity(PiwigoResponse.class);
```

Will send a request url: `https://<hostname>/ws.php?format=json&method=pwg.images.getInfo&image_id=339`

And result in an exception: 
```
org.glassfish.jersey.message.internal.MessageBodyProviderNotFoundException: MessageBodyReader not found for media type=text/plain;charset=utf-8, type=class <package>.PiwigoResponse, genericType=class <package>.PiwigoResponse.
```

mingan666:
Fix #1847 

Changes mediatype to application/json

### DIFF
--- a/admin/themes/default/template/intro.tpl
+++ b/admin/themes/default/template/intro.tpl
@@ -193,7 +193,7 @@ let str_size = "";
           {if $SIZE != 0}
           {assign var='SIZE_IN_UNIT' value=$SIZE/$ACTIVITY_CHART_NUMBER_SIZES * 5 + 1}
           {assign var='OPACITY_IN_UNIT' value=$SIZE/$ACTIVITY_CHART_NUMBER_SIZES * 0.6 + 0.2}
-          <div id="day{$WEEK_NUMBER}-{$DAY_NUMBER}" style="height:{$SIZE_IN_UNIT}vw;width:{$SIZE_IN_UNIT}vw;opacity:{$OPACITY_IN_UNIT}"></div>
+          <div id="day{$WEEK_NUMBER}-{$DAY_NUMBER}" style="height:{$SIZE_IN_UNIT}vw;width:{$SIZE_IN_UNIT}vw;"></div>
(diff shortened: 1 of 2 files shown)


plegall:
Hi @mingan666 

Thank you for this pull-request. Why don't we immediately merge it, as it seems so obvious it's the right fix? Because we don't want to break third party applications which would rely on the current `text/plain` mediatype.

Related to #1107 #406 or #448

The question is for me : how to make things right with `application/json` but not break old third party apps relying on `text/plain`? Then you understand why it's not as easy as it seems :-/

plegall:
(I closed this PR by mistake, sorry)

mingan666:
Thanks for the explanation.

I understand your concern, but at the same time the current bug is causing problems for 3th party libraries that are not able to recognize the response as json.
Would it be possible to add an additional query parameter requesting the proper mediatype?

Not that good at php so I'm not sure how to implement such a solution...

mingan666:
Quick look suggest it would be possible to add an additional format into `include/ws_init.inc.php` and `tools/ws.htm`.
As well as an additional class that could extend `include/ws_protocols/json_encoder.php`

Not the finast solution, but it gives users that want a correct mediatype that option, without breaking exiting integrations.
As I wrote above, not that good att php so I might be missing other parts that might need tweaking as well...

PR #1850, opened 2023-01-10
Fixes #1254 fix quote escaping in password

### LINKED ISSUE #1254: Password with quote (') character
Hi,

I'm using piwigo with the ldap plugin. One of my users have a quote (') in its password. Piwigo badly handles it by quoting it and not removing the quote. I did some tests and see that "let'sgo" is stored as "let\'sgo" in $_POST['password'] (due to quote encoding) and it seems to never be decoded (same for $_POST['username'] even if I did not see a username with a quote)

I fixed my installation with this patch :
```php
diff --git a/piwigo/identification.php b/piwigo/identification.php
index 9059ef2..84a2ac1 100644
--- a/piwigo/identification.php
+++ b/piwigo/identification.php
@@ -60,7 +60,7 @@ if (isset($_POST['login']))
     $redirect_to = isset($_POST['redirect']) ? urldecode($_POST['redirect']) : '';
     $remember_me = isset($_POST['remember_me']) and $_POST['remember_me']==1;
 
-    if ( try_log_user($_POST['username'], $_POST['password'], $remember_me) )
+    if ( try_log_user(stripcslashes($_POST['username']), stripcslashes($_POST['password']), $remember_me) )
     {
       // security (level 2): force redirect within Piwigo. We redirect to
       // absolute root url, including http(s)://, without the cookie path,
```
I found strange that piwigo use magic quoting (doing it manually if it is disabled on the server) instead of properly managing input and doing quoting when required (SQL requests, ...)

Regards
   Vincent

zag-adka:


include/functions_user.inc.php
```diff
@@ -1168,7 +1168,7 @@ function pwg_login($success, $username, $password, $remember_me)
 ;';
 
   $row = pwg_db_fetch_assoc(pwg_query($query));
-  if (isset($row['id']) and $conf['password_verify']($password, $row['password'], $row['id']))
+  if (isset($row['id']) and $conf['password_verify'](stripslashes($password), $row['password'], $row['id']))
```
nikrou, line 1171: Very bad idea ! You should use mysql_real_escape_string at least

zag-adka: No, because its doesn't generate a SQL request. The given password is passed to the hash function and compared to the hashed password that is stored in the database. The hash function has to get the password without slashes.

But maybe you prefer to put the stripslashes in password_verify function.

PR #1853, opened 2023-01-21
Introduce $conf['picture_page_banner'] to hide page banner on picture page

erAck:
The configurable page banner can take quite some precious space
from individual picture pages and may not be wanted there.
$conf['picture_page_banner'] = false;
in local config can disable it.

### DIFF
--- a/picture.php
+++ b/picture.php
@@ -1029,6 +1029,10 @@ if ($conf['picture_menu'] AND (!isset($themeconf['hide_menu_on']) OR !in_array('
   include( PHPWG_ROOT_PATH.'include/menubar.inc.php');
 }
 
+if (isset($conf['picture_page_banner']) && !$conf['picture_page_banner'])
+{
+  $page['page_banner'] = '';
+}
 include(PHPWG_ROOT_PATH.'include/page_header.php');
 trigger_notify('loc_end_picture');
 flush_page_messages();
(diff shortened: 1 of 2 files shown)


picture.php
```diff
@@ -1029,6 +1029,10 @@ function default_picture_content($content, $element_info)
   include( PHPWG_ROOT_PATH.'include/menubar.inc.php');
 }
 
+if (isset($conf['picture_page_banner']) && !$conf['picture_page_banner'])
```
plegall, line 1032: you don't need to check is `isset`. It's always set since you have it in `include/config_default.inc.php`

plegall:
I think it's a good idea. But I would think it's an even better idea to have it manageable on page [Admin>Config>Options>Display]. It's much more complex to implement be seems better to me. What do you think?

picture.php
erAck: > you don't need to check if `isset`. It's always set since you have it in `include/config_default.inc.php`

Right. Probably was a habit as some time ago I fiddled around with (other) config settings in admin.php and there was some situation when config wasn't loaded yet.


> I think it's a good idea. But I would think it's an even better idea to have it manageable on page [Admin>Config>Options>Display]. It's much more complex to implement be seems better to me. What do you think?

Sure, feel free :-)  I have it just as a quick&dirty hack in my gallery. I also don't know how another solution might interact with different themes (seems the banner isn't available in all), "works for me" ...

PR #1857, opened 2023-02-02
Added two Finnish translations

zikkee:


ddtddt:
Hi :-)

For translation please use [tool](https://piwigo.org/translate/)

Register ans after I open Right

PR #1885, opened 2023-03-26
Fix additional format with application/json

mingan666:
Fix https://github.com/Piwigo/Piwigo/issues/1847

Adds additional format for media type application/json

@plegall Hopefully this fix might be more acceptable then the previous attempt (https://github.com/Piwigo/Piwigo/issues/1847)

### DIFF
--- a/include/ws_init.inc.php
+++ b/include/ws_init.inc.php
@@ -58,6 +58,10 @@ if (!is_null($responseFormat))
       include_once(PHPWG_ROOT_PATH.'include/ws_protocols/json_encoder.php');
       $encoder = new PwgJsonEncoder();
       break;
+    case 'jsonalt':
+      include_once(PHPWG_ROOT_PATH.'include/ws_protocols/jsonalt_encoder.php');
+      $encoder = new PwgJsonEncoder();
+      break;
     case 'xmlrpc':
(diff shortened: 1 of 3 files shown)


mingan666:
I see I forgot to update after my fixes. I'll try to find some time to update my patch later today...

mingan666:
Think I fixed the issues. Please let me know what you think.

klaoun:
for example when I want to add categories I put the name in "name" --> json(application/json) and I invoke
it gives me the result 

"Cannot process your request. Unknown response format.
Request format: rest Response format: jsonalt
PwgServer::__set_state(array(
   '_requestHandler' => "

but why is this

THANKS

mingan666:
I don't really understand your question.
For me the following works:

```
https://<hostname>/ws.php?format=jsonalt&method=pwg.images.getInfo&image_id=11

{
    "stat": "ok",
    "result": {
        "id": 11,
        "file": "DSC08817.jpg",
...
```

and after login:

```
https://piwigo.rutgers.se/ws.php?format=jsonalt&method=pwg.categories.add&name=test

{
    "stat": "ok",
    "result": {
        "info": "Album added",
        "id": 49
    }
}
```
However I've given up and moved to another gallery implementation due to the bad response on both github and the forum from those that understand the system. It's to hard to fix the integrations I need.

Issue #1938, opened 2023-07-08
php-8.2 issue with Piwigo 13.8.0


AStcherbinine:
Hello,
I just upgraded my server from Debian 11 to Debian 12, leading the PHP version to be upgraded from 7.4 from 8.2, and now I have all these issues printed on the web page when I try to access my Piwigo instance (copied below).
Does anyone know where it comes from and if there is a workaround to fix it?

Server: Debian 12.0
Piwigo version: 13.8.0
PHP: 8.2.7
SQL: 10.11.3-MariaDB-1 Debian 12

Thank you!

```
Deprecated: Creation of dynamic property Smarty_Internal_Extension_Handler::$registerPlugin is deprecated in /srv/www/piwigo/include/smarty/libs/sysplugins/smarty_internal_extension_handler.php on line 182

Deprecated: Creation of dynamic property Smarty_Internal_Extension_Handler::$registerFilter is deprecated in /srv/www/piwigo/include/smarty/libs/sysplugins/smarty_internal_extension_handler.php on line 182

Deprecated: Creation of dynamic property Smarty_Internal_Extension_Handler::$append is deprecated in /srv/www/piwigo/include/smarty/libs/sysplugins/smarty_internal_extension_handler.php on line 182

Deprecated: Creation of dynamic property Smarty_Internal_Extension_Handler::$getTemplateVars is deprecated in /srv/www/piwigo/include/smarty/libs/sysplugins/smarty_internal_extension_handler.php on line 182

Deprecated: Creation of dynamic property Smarty_Internal_Extension_Handler::$unregisterFilter is deprecated in /srv/www/piwigo/include/smarty/libs/sysplugins/smarty_internal_extension_handler.php on line 182

Deprecated: Creation of dynamic property Smarty_Internal_Extension_Handler::$clearAssign is deprecated in /srv/www/piwigo/include/smarty/libs/sysplugins/smarty_internal_extension_handler.php on line 182

Deprecated: Creation of dynamic property CssAtMediaStartToken::$MediaTypes is deprecated in /srv/www/piwigo/include/cssmin.class.php on line 34

Deprecated: Creation of dynamic property CssAtMediaStartToken::$MediaTypes is deprecated in /srv/www/piwigo/include/cssmin.class.php on line 34

Deprecated: Creation of dynamic property CssAtMediaStartToken::$MediaTypes is deprecated in /srv/www/piwigo/include/cssmin.class.php on line 34

[...]

PR #1944, opened 2023-07-20
Update cssmin.class.php

osworx:
Fix php 8.2.x error:
Deprecated: Creation of dynamic property CssAtMediaStartToken::$MediaTypes is deprecated in /srv/www/piwigo/include/cssmin.class.php on line 34

See https://github.com/Piwigo/Piwigo/issues/1938

While I do not really understand why this old library is further used .. ?!


plegall:
Can you elaborate on the way you fixed the problem? (the diff is not readable).

> While I do not really understand why this old library is further used .. ?!

Because we didn't find any replacement.

osworx:
Fixing the error:

(had to rework the whole file to find the proper position ..):

1. find (in the minified version): **class CssAtMediaStartToken extends aCssAtBlockStartToken {**
2. add directly after: **public $MediaTypes = [];**

Reg. the old version (and any new replacement) - maybe one of them:
A. https://github.com/matthiasmullie/minify - php (actively developed, last release June 8th 2023)
B. https://github.com/parcel-bundler/lightningcss - rust (also actively developed, last release July 2023)
C. https://github.com/tdewolff/minify - go (last release June 2023) / also HTML, php, js ..

Of course the old package could be reworked, but when newer packages are exisiting?

plegall:
Concerning a replacement, only https://github.com/matthiasmullie/minify could be candidate, since it's the only one in PHP and we don't plan to rewrite Piwigo in Rust or Go :-)

Concerning the fix, I thought you had "un-minified" the cssmin.class.php file and I thought it could be good idea to have its unminified version in Piwigo source code files (to make things easier to update next time). We could have `cssmin.class.php` as un-minified version and `cssmin.class.min.php` as minified version and use this last one (even if I think it's a completely minor performance improvement compared to including the un-minified version).

plegall:
The problem with this pull-request is that the whole file is modified. I can't see what has really changed and maybe there is much more modification [... 5 minutes later...] I have now "formatted" `ccsmin.class.php` before and after your change and here is the diff:

```
% diff cssmin-before.class.php cssmin-after.class.php
2421a2422
>   public $MediaTypes = [];
```
Much more readable, isn't it?

I merge this pull-request, but I have also created a new issue #1948 to study the replacement of `cssmin.class.php` for a future release of Piwigo.

PR #1961, opened 2023-08-18
Use smarty 4.3.2

gotztibor:
Version upgrade of smarty to 4.3.2 for better PHP 8.2 compatibility (deprecated dynamic properties)


osworx:
Just 4 my info: why array() and not the modern usage (since php 7.x) [] ?

plegall:
@gotztibor why are you mixing update of Smarty and a change in `cssmin` ?

PR #1991, opened 2023-09-24
Fix warning in pwg.categories.php

mistic100:
When calling "pwg.categories.getImages" with a unknow "cat_id" there is PHP notice because the $total_images variable is not defined.

Example : https://galerie.strangeplanet.fr/ws.php?format=json&method=pwg.categories.getImages&cat_id=2

This is a quick fix, perhaps the endpoint should fail when the category does not exist ?

### DIFF
--- a/include/ws_functions/pwg.categories.php
+++ b/include/ws_functions/pwg.categories.php
@@ -22,6 +22,7 @@ function ws_categories_getImages($params, &$service)
 
   $images = array();
   $image_ids = array();
+  $total_images = 0;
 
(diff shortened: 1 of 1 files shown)


plegall:
I merge it, but yes, a check on existing `category_id` would be better. I'm going to add it.

plegall:
Actually, as you can provide no `cat_id` to `pwg.categories.getImages`, your fix is necessary. Independantly of the additional check I've added in 4c734bedd80ff3c3f481c1a2a25990916105e0ae

PR #1999, opened 2023-09-24
Update album_selector.inc.tpl

osworx:
bugfix for:

PHP Warning:  Undefined array key "show_root_btn" in _data/templates_c/....file.album_selector.inc.tpl.php on line 72

### DIFF
--- a/admin/themes/default/template/include/album_selector.inc.tpl
+++ b/admin/themes/default/template/include/album_selector.inc.tpl
@@ -25,7 +25,7 @@
       <span>{$title}</span>
     </div>
 
-    {if $show_root_btn}
+    {if isset($show_root_btn)}
     <label class="head-button-2 put-to-root">
       <p class="icon-home">{'Put at the root'|@translate}</p>
     </label>


plegall:
I reproduce this bug on photo edit page, but this is not the right way to fix it. If `show_root_btn` is set to false, this code will show the root btn. Which is obviously not what we want. The `show_root_btn` is variable in sub-template `album_selector.inc.tpl` and is provided during include, like this (in `themes/default/template/include/search_filters.inc.tpl`)

```
  {include file='admin/themes/default/template/include/album_selector.inc.tpl'
    title={'Search in albums'|@translate}
    searchPlaceholder={'Search'|@translate}
    show_root_btn=false
    api_method='pwg.categories.getList'
  } 
```

You also have the `title` variable and you don't make an `isset($title)` in `album_selector.inc.tpl`. We'd better deal with default value like it's done for `$load_mode`:

```
{if !isset($show_root_btn)}{$show_root_btn=false}{/if}
```

(I've tested, works fine) What do you think?

osworx:
Well, I do not know the Piwigo system so good as you ;)
Finally it's your decision how to deal with such messages, undefined variables, etc.

Any `issset(..)` is better than nothing, but it's not the way I work(ed) until today (in those systems all variables have to be defined/initiated before further usage (because of the fact that undefined variables but called later somewhere, a security risk can be).

But as written, when I know how you want to have things fixed, it's easier for me to make proper PRs.

p.s.: to be honest, the Piwigo code is sometimes "hard to read", I used to work with classes and functions all the time.
And Smarty was also not my friend before Piwigo ..

plegall:
> Well, I do not know the Piwigo system so good as you ;) Finally it's your decision how to deal with such messages, undefined variables, etc.

Of course I know Piwigo code quite well, but that doesn't mean I can't "explain" why this pull-request can't be merged "as is". I want to point out that dealing with a pull-request requires some time to test, to think about the right way to fix the issue and so on. I could have simply pressed the "rebase and merge" button, but only to get a new, less obvious, bug in the user interface.

> Any `issset(..)` is better than nothing

I don't agree with that. If it's only to hide a bigger issue, better keep the error and correctly fix it.

>, but it's not the way I work(ed) until today (in those systems all variables have to be defined/initiated before further usage (because of the fact that undefined variables but called later somewhere, a security risk can be).

This is not how PHP works. For years, using undefined variables in the template was not an issue. With PHP 8 it has become an issue and now we have to "fix" template files (but they were working perfectly before that).

> p.s.: to be honest, the Piwigo code is sometimes "hard to read", I used to work with classes and functions all the time. And Smarty was also not my friend before Piwigo ..

Not sure Piwigo would be easier to read in a full object-oriented coding style ;-) Anyway, this is something we can't really change (or we would lose all plugins compatibility).

I've implemented the fix as described earlier and I close this pull-request. Thank you again for making this bug obvious, I had not seen it. See #2009

osworx:
Generally: it does no matter how you solve bugs, important is, that they are solved.
That's why I wrote 

> Any issset(..) is better than nothing

For sure, not the best option (fully agree with you), but a solution.

And whenever I can help (test, etc.) you can count on me.
What I have meant with the readabilty was that sometimes the code is a mix of functions, calls and other functions.
A OOP would be nicer, but as you say that will break the plugin compatibility.
Maybe a goal for version 15 ?
Because what I have seen until now, many plugins are not maintained nor really ready for php 8.2.x - had to make also at some PRs.

PR #2010, opened 2023-09-29
Update cat_modify.php

klaoun:
this page is compatible php 5 to php 8.3
try removing the comma it's ok


klaoun:
hello it 's corrected #2080 
thx

PR #2012, opened 2023-09-30
Update search_filters.inc.tpl

osworx:
Missing translation.

btw: what for you are using the @ before translate??


plegall:
the @ is useless with recent version of smarty

osworx:
Thx for answering.
I saw that at the Smarty documentation, that's why I wondered myself why it's still used.

PR #2121, opened 2024-02-27
Fix nl_NL translation

erAck:
Some-one|thing mistook the u in %u for "you" and translated the
specifier to %jouw where %j leads to
Fatal error: Uncaught ValueError: Unknown format specifier "j"

See https://piwigo.org/forum/viewtopic.php?id=33492

Already since

    commit 6011ae152d94cad2cdae3776fa5d1584ad2e7bc2 (tag: 2.9.0RC1)
    CommitDate: Fri Mar 31 15:55:26 2017 +0200

        update language directory from the "translation" branch


erAck:
This should also be cherry-picked to the `14.x` branch.

Issue #2124, opened 2024-03-01
Typo: When you want to sort your albums there is twice the ability to sort A-Z


MaikelvanGorkom:
While the second is sort order Z-A

MaikelvanGorkom:
Fixed it in #2125

PR #2125, opened 2024-03-01
Fix for #2124: en_GB had a typo in sort order.

MaikelvanGorkom:
This to make the line according to what it does.


plegall:
I have also reported this change on branch `translation` with piwigo.org/translate

PR #2126, opened 2024-03-03
Correction for  PHP Deprecated:  explode()

Hood-fr:
More robust syntax if $_GET['section'] turns out to be null, avoid messages as PHP Deprecated:  explode(): Passing null to parameter #2 ($string) of type string is deprecated


klaoun:
Hello
This code is compatible PHP 7.0 ?

klaoun:
Hi

is it possible to write this code -->

if `$sections = explode('/', $_GET['section'] ?? '');` is good then that otherwise this code
   `$sections = explode('/', $_GET['section']);`      ???   or these 2 ?

thanks

PR #2183, opened 2024-07-12
Fix typo. "Abum" to "Album

asandikci:


patman-cp (approved):
Not a piwigo developer but this looks good to me.

Note that you can see they typo if you go to the Administrator -> Albums  Properties.

Gazook89 (approved):
Also not a developer but also agree it looks good. It's still an issue and this fixes it, though I see that upstream changes need to merged into it and conflicts resolved. 

From live site:
image

PR #2214, opened 2024-08-30
Fix "total_count" type on some WS methods

mistic100:
Working on a JS client for the WS I noticed some "total_count" field that were returned as strings.

Also the "total_count" on the ws_users_getList method was not at the same location as other paginated methods.

### DIFF
--- a/admin/themes/default/js/user_list.js
+++ b/admin/themes/default/js/user_list.js
@@ -2087,12 +2087,12 @@ function update_user_list() {
                 // console.log(data.message);
                 return;
             }
-            total_users = data.result.total_count;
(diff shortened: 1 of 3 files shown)


plegall:
Hi @mistic100 

Don't you fear some 3rd party apps rely on the fact that `total_count` is a string? and we would "break" them by forcing to an int?

mistic100:
I don't :-)

Dynamically typed languages won't feel the difference. Statically typed languages most commonly use dedicated JSON parsers which are resilient to this kind of casting (at least it the case of Jackson for Java, nlohmann/json for C++).

Of course if someone did `total_count.slice(1, -1)` it will break but.... yeah.

PR #2216, opened 2024-08-30
Draft: Add "pwg.getConfig" WS method

mistic100:
**To be discussed.**

I try to start working on my PWA based on Piwigo WS and am in need of some configuration parameters of the server. 

- I personally only need behavior parameters, thus everything directly related to the UI is not exposed, though the new method could have a parameter to choose what to get (`general`, `ui`, `all` ... ?)

- These parameters should not have any security impact, but if more are exposed, a test of the user role will be needed.

- I actually wonder what is `gallery_locked` ? Was it removed form the admin page or is it only setted during updates ?


plegall:
While I find the idea to be able to fetch the configuration with an API call, I'm absolutely not comfortable to provide such information for "anyone who ask". At the very least it should be "admin only". I would even prefer a "webmaster only" filter.

mistic100:
In my case, the properties I listed **need** to be "public". Those are only properties related to the UI, how many comments to display, is the rating widget enabled, etc. If it is for admin/webmaster only, it is useless.

PR #2298, opened 2024-12-25
added support for docker

shubhamgarg21:


jheinitz:
Nice. Finally we will have official docker support for Piwigo. Thanks.

Issue #2360, opened 2025-04-25
New visual for jgrowl pop-up


Martin-Raby:
Change the jgrowl pop-up visual
![Image](https://github.com/user-attachments/assets/6141f64b-3073-4e3c-80a7-1f33fd20bf67)

PR #2363, opened 2025-04-29
Change in the jGrowl popup notification visual

Martin-Raby:
#2360
- New CSS to change the style
- New JS to add the icon


HWFord (changes requested):
Lack of padding underneath the texte
image

PR #2432, opened 2025-11-07
xss

hugbubby:


hugbubby:
ah fuck meant to send this to my own thing to test something out, this is literally an xss

PR #2487, opened 2025-12-19
imagick7 expects input file before options

camillehuot:
the old code does not work with ImageMagick 7


plegall:
I'm pretty sure we have tested Piwigo 16.0.0 with ImageMagick 7 (and it was working, of course). This is why we have added a change of command line from `convert` to `magick`.

In your case, the command line completely fails? (before your proposed change, I mean)

PR #2562, opened 2026-05-31
security: add allowed_classes => false to unserialize() in safe_unserialize(), picture.php, and template.class.php

XananasX7:
## Summary

Three `unserialize()` call sites process configuration values from Piwigo's database (`piwigo_config` table) without specifying `allowed_classes`.

### `safe_unserialize()` — `i.php`
Despite the name, this function only checks whether the input is a string before calling `unserialize()`. It doesn't restrict which PHP classes can be instantiated. This function is used throughout the codebase. An attacker who can write to the config table can inject a serialized object payload.

### `picture.php` — `picture_informations`
Deserializes the `picture_informations` configuration value, which is expected to be a plain array of display flags (booleans/integers).

### `template.class.php` — `extents_for_templates`
Deserializes the `extents_for_templates` configuration value, which is expected to be a plain array of template extension paths (strings).

## Fix
Added `['allowed_classes' => false]` to all three call sites. The stored values are plain PHP arrays — no objects are expected. Zero behavior change for valid data.

### DIFF
--- a/i.php
+++ b/i.php
@@ -369,7 +369,7 @@ function safe_unserialize($value)
 {
   if (is_string($value))
   {
-    return unserialize($value);
+    return unserialize($value, ['allowed_classes' => false]);
   }
   return $value;
 }
--- a/include/template.class.php
+++ b/include/template.class.php
@@ -181,7 +181,7 @@ class Template
 
     if (!defined('IN_ADMIN') and isset($conf['extents_for_templates']))
     {
-      $tpl_extents = unserialize($conf['extents_for_templates']);
+      $tpl_extents = unserialize($conf['extents_for_templates'], ['allowed_classes' => false]);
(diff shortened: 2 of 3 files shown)


klaoun2:
hi what php version ???

XananasX7:
Hi @klaoun2! The `allowed_classes` option for `unserialize()` was added in PHP 7.0, so this is compatible with any PHP version Piwigo currently supports. No version requirement change needed.

plegall:
Hi @XananasX7 

Can you make the following change : where `unserialize` is used, replace it with `safe_unserialize`. Then we would have to deal with `allowed_classes` only in the `safe_unserialize` function.

XananasX7:
$tpl_extents = unserialize($conf['extents_for_templates']);

Issue #2536, opened 2026-03-07
PHP warnings/errors.


rewolff:
A friend came to me with the problem that his piwigo installation wasn't working after an upgrade... So we looked into that and seemed to have fixed it with adding the column "preferences" to the user_infos table. 

So then I asked: Can you give me a "getting started with piwigo" tutorial. So I installed it on my own computer and we started importing photos. 

After a while we got: 
```
 PHP Warning:  Array to string conversion in /var/www/html/piwigo/include/functions_category.inc.php on line 162, referer: ... piwigo/admin.php?page=batch_manager&display=30
 PHP Fatal error:  Uncaught mysqli_sql_exception: Unknown column 'Array' in 'WHERE' in /var/www/html/piwigo/include/dblayer/functions_mysqli.inc.php:132
```
which happens to be the exact same error that my friend was originally complaining about. 

This is how I fixed it: 
```
function get_cat_info($id)
{
  $id = is_array($id) ? $id[0] : $id;
  error_log("in get_cat_info: id is: " . print_r($id, true));

  $query = '
SELECT *
  FROM '.CATEGORIES_TABLE.'
  WHERE id = '.$id.'
```
Apparently get_cat_info is now called with an array instead of a single int as the "$id". 

piwigo 16.3.0 (feb 24? downloaded zipfile from piwigo site). 
Ubuntu 24.04,  
php 8.3 

My friend thought it was "version problems" so he downgraded piwigo to 15.x and php from 8.4 to 8.3. (on his ubuntu 25.10 system.)

tomchiverton:
This happens here on Fedora Core 43, PHP 8.4, latest Piwigo 16.4.0 :( Every single time I assign to album via batch manager.

Adding `  $id = is_array($id) ? $id[0] : $id;` right before `$query ... ` fixes it.

tomchiverton:
MR : https://github.com/Piwigo/Piwigo/pull/2584

PR #2584, opened 2026-09-20
Fix https://github.com/Piwigo/Piwigo/issues/2536

tomchiverton:


plegall:
Hi @tomchiverton 

Your PR is "my functions receives an array instead of an int, let's take the first int". I don't like that because I need to understand why/who sends this array instead of a single int.

Your PR made me investigate and I have commited a more robust solution, see 8b63248aa4ea8e71c9f1f4ecd85773b60ece7b44 : thank you for forcing me to work on a solution 👍

tomchiverton:
Thanks for a better fix.

Will this make a release soon, as seems a core feature is broken?
-- 
†øღ

